PR #8, opened 2022-02-26
Add Per World Configuration Guide

sulu5890:
designed as a precursor to the anti-xray guide and other future documentation. This is very commonly misunderstood so official docs on how it works would be good.


docs/paper/how-to/per-world-config.md
```diff
@@ -0,0 +1,95 @@
+# Per World Configuration
+
+One of the most powerful yet least understood features of the Paper configuration is setting
+configuration options per world. While not every config option can be set per world, everything
+under `world-settings` in either `paper.yml` or `spigot.yml` can be configured differently on a per
+world basis.
+
+## Default Values
+
+The only world generated by out of the box is `default`. Any configuration option set here will
```
chickeneer, line 10: I think there is an extra "by" here. 

> generated by out of the box

docs/paper/how-to/per-world-config.md
```diff
@@ -0,0 +1,95 @@
...
+| `spawn-limits.axolotls`                   | `5`    | `5`          | `5`           | `8`            |
+| `spawn-limits.underground_water_creature` | `5`    | `5`          | `5`           | `5`            |
+| `spawn-limits.water_creature`             | `5`    | `5`          | `5`           | `5`            |
+| `spawn-limits.water_ambient`              | `20`   | `20`         | `20`          | `20`           |
+| `keep-spawn-loaded`                       | `true` | `true`       | `true`        | `false`        |
+
+Notice that `world_the_end` was never specified in this configuration. Because of this, it inherits
+all the configuration options from the `default` section. Additionally, `keep-spawn-loaded` was only
+disabled in `resource_world`. Because in the `default` section, `keep-spawn-loaded` is set to
+`true`, all worlds except for `resource_world` will have `keep-spawn-loaded` enabled.
```
chickeneer, line 95: These last two sentences are awkward to read to me and partially redundant.
I would simplify it to something like 

> Additionally, `keep-spawn-loaded` was only disabled in `resource_world` because in the `default` section, `keep-spawn-loaded` is set to `true`.

PR #16, opened 2022-03-06
reorganization and style changes

sulu5890:
attempts to improve the colours and such a bit, feedback greatly appreciated

~~Todo: code block titles are a weird colour
Screenshot_20220306-005117~~


sulu5890:
I'm happy enough with this. can always change more later, but at least it is more legible now.

PR #25, opened 2022-03-09
Fix header for timings category

NoahvdAa:
Because the header was set to `##` instead of `###`, some settings were unintentionally added as children of the timings category:
image


sulu5890 (approved):
Good catch

PR #29, opened 2022-03-15
Document side effect of disable-ice-and-snow

NoahvdAa:
Noticed this while researching PaperMC/Paper#7605.

### DIFF
--- a/docs/paper/reference/paper-per-world-configuration.md
+++ b/docs/paper/reference/paper-per-world-configuration.md
@@ -228,7 +228,7 @@ global config, see the [Paper Global Configuration Reference](paper-global-confi
 ### disable-ice-and-snow
 
 - **default**: false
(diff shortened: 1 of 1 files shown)


Owen1212055:
I assume this also effects them being able to be filled up by snow as well?

Machine-Maker:
how long has rain filled up cauldrons? does this config pre-date that? if so, I feel like the solution is to make that config *not* affect this.

NoahvdAa:
According to [the Minecraft wiki](https://minecraft.fandom.com/wiki/Cauldron), rain filling cauldrons was added in Minecraft 1.3.1.

Machine-Maker:
k, that's a lot older than I thought. guess it can stay. just seems weird 🤷

sulu5890 (approved):

PR #30, opened 2022-03-15
Rename Reference categories Configuration

sulu5890:
Initially done at the request of @kennytv as the (Paper) Reference category currently contains just the configuration file.

My plans for the category in the future were:
* Built-in commands overview
* System properties
* possibly other "reference" type stuff I haven't thought of yet

which are also sort of configurations in a sense.

Opened primarily for opinion as it's not something I am so keen on, but at least right now, it does make more sense on the sidebar.

**TODO:**

There aren't any redirects implemented for the changed URLs. Doing so is a must should this be merged. Unfortunately, GitHub pages does not support proper redirects, so more to be investigated there. ugly solution is just [@docusaurus/plugin-client-redirects](https://docusaurus.io/docs/api/plugins/@docusaurus/plugin-client-redirects)


Machine-Maker:
I don't really think commands fall under the "Configuration" title.

sulu5890:
yeah, as I think about this more and more especially with the config migration it doesn't really make sense.

PR #33, opened 2022-03-26
add structured data for search, support adding custom html tags to header/body in config

sulu5890:


sulu5890:
looks to work correctly

test result

PR #42, opened 2022-03-30
Breadcrumb links to all pages, category indexes

sulu5890:
Fixes an issue @MiniDigger noticed with google crying about invalid breadcrumbs.

This is intentional on the end of Docusaurus, ref https://togithub.com/facebook/docusaurus/pull/6932. Possibly a better solution is just to remove the last breadcrumb. I'm not sure, as there is good rationale provided there.


sulu5890:
Still don't love this but it seems like the best solution I've found and is at least okayish. Don't see anything else being better.

PR #45, opened 2022-04-03
Removed Event interface from custom velocity event example

4drian3d:
Actually the example contradicts itself by saying `You'll notice that your events don't need to extend or implement anything. They just work.` being that the example implements an interface and the `Event` interface never existed and confuses many people thinking that they have to implement this interface


sulu5890 (approved):
The velocity docs are in a bit of a limbo state right now with the three different versions. That'll hopefully be worked out soon to make everything more clear and then more revision can be done to make sure everything is up to date and makes sense. Regardless, looks good. Thanks.

PR #48, opened 2022-04-04
Refactor for multiple docs plugin instances and sidebars

sulu5890:
Also see #50 for an alternative idea.

Refactor to separate the different projects more, both in terms of the sidebar and how they are dealt with internally. It does not break URLs, though, outside of the `papermc` group, which has been moved into `common`.

This is mostly just experimental/a concept. I'm still pretty much 50/50 as to if this is better or not. Mostly open for feedback. This isn't complete.

Some advantages that this does provide, more for the future:

Should we want to implement versioning, Paper and Velocity could be versioned differently, while Waterfall/Common wouldn't need to be versioned at all.
Localization would also be cleaner to implement, should we want that.

Something applicable now is changing all of the sidebars to be independent. You will now only see the sidebar for the project you are browsing. This is doable without different `content-docs` plugin instances for each set of docs, but I figured why not just get that out of the way now, so it is less pain should we want to version later down the line.

A limitation of this with the current architecture of Docusaurus, you can't use markdown links between docs in different sections. You instead need to link to its slug (see the java update guide link changes). This has no change in the built site but is slightly more work in writing docs. This would be solved by something like [this proposal](https://togithub.com/facebook/docusaurus/issues/6370). The same applies to `doc` sidebar link types. A `link` type would be required instead.

Todo:

1. Create a proper `getting help` section (under common) and add to the top of each sidebar
2. Use search facets to limit search to the project you are currently viewing (with a toggle?) and common. Homepage search or search on `common` will still display all results.
3. Other things I am forgetting now, I know I had way more than this when I started to type out the comment. I'll update this


config/common.ts
```diff
@@ -0,0 +1,25 @@
+import { SidebarsConfig } from "@docusaurus/plugin-content-docs";
+
+const sidebarType: SidebarsConfig = {
+  sidebar: [],
+};
+
+export const sidebar: typeof sidebarType.sidebar = [
+  {
```
sulu5890, line 8: This is completely awful. I don't know how to make it better though (and am pretty new to typescript). The type needed isn't exported, so this hack is used instead.

sulu5890:
After conversation on discord, this isn't preferred.

Machine-Maker:
I think there shouldn't be a global sidebar. There are few cases where, in one session, one is jumping about between different platforms' admin docs and developer docs. I think there should be 3 sidebars, one for each platform, and then for each platform, there is two top level sidebar entries, one for server admins, one for developers. and a landing page that just has a small blurb about each platform and links to the admin/dev page for that platform

sulu5890:
Made a bunch of changes to sorta match that. Now:

Four total sidebars:
1. Paper
2. Velocity
3. Waterfall
4. Common

`paper` and `velocity` sidebar are split into `admin` and `dev`, while waterfall and common sidebars are unified (waterfall does not have a dev section).

With this also comes wayyyy more URL changes, so more redirects to add once this is done.

Machine-Maker:
Looks good. couple inconsistencies with sidebar item naming and such, but the structure looks good. I think the Welcome pages need to be renamed or have a different name template, `Welcome to Blah` seems too long, and repetitive, idk what the best solution is, just `Welcome` or `Paper/Waterfall/Velocity`. I think the welcome pages should also have big button links to each top and 1-level below section of their sidebars, or at least if the sidebars can be expanded on page load.

sulu5890:
Yeah, buttons would be nice. I had investigated this when initially implementing them, using the built-in [DocCard](https://github.com/facebook/docusaurus/blob/main/packages/docusaurus-theme-classic/src/theme/DocCardList/index.tsx)s - however, it's not feasible to autogenerate this as global sidebar state isn't available here.

So essentially, it'd either have to be hacked together with some weird post-processing or done manually, which are both far from ideal. I'll further investigate the post-processing possibility as it would be a significant improvement.

sulu5890:
I've gone for a different approach and instead added these link items to each category (admin/dev) and made that the main landing page from the navbar. An issue I've noticed is that the `activeBaseRegex` does not seem to correctly apply for the dropdowns themselves (although for its items, it works perfectly). I've also yet to update the home page.

sulu5890:
Happy enough with the organization here now. Still have to work on some UI issues, but that can be done as a separate PR.

PR #49, opened 2022-04-05
Remove chunk-tasks-per-tick

HexedHero:
It got removed in the Tuinity merge here https://github.com/PaperMC/Paper/commit/7d10cdea0358854138d0d0ada5e74c58c688dea7#diff-cabcb269ec3936b8c8cd99cfe30492bc7aafc3a835fae86fda89b7543bc1d5f4L21


sulu5890 (approved):
Thanks. Should set up a script to check this (adds/deletes vs what is documented) automatically. Will do that shortly.

PR #53, opened 2022-05-07
Update alt-item-despawn-rate docs

NoahvdAa:
https://github.com/PaperMC/Paper/commit/2ab0258b5ab985da6d8fe3728c994b325c08ada0


sulu5890 (approved):
Looks good, thanks.

I'd like to in the future, now that this is fully standard, rather than labeling each individually create a "types" table at the top detailing all custom types used (likely excluding one-off types, although I'm not 100% on that yet), and then just add a type for each config option. This would include non-custom types as well.

Main issue with this is how the config docs are split across two sections which contain the same types. I'm not sure if there is any community standard way to handle partials/replacement for markdown. So that's something to look into both for here and other smaller places.

PR #56, opened 2022-05-26
Fix bamboo indent

### LINKED ISSUE #55: Bamboo min value not listed under bamboo
Bamboo minimum grow height value is not listed under the bamboo list item
image

Doc94:
This fix #55 

PD: Thanks windows for take 30min for fix a gpg error...


sulu5890 (approved):
Thanks

PR #61, opened 2022-06-08
Clarify navigation by marking current sidebar page with activity bar

sulu5890:
I'm not quite sure I like the look of this right now, especially as it pushes the actual text on the sidebar over a bit. This was to prevent word wrapping changes when selecting a sidebar item. I'm not sure that it's the correct way to do that. There might be something better.

unwanted behaviour:

https://user-images.githubusercontent.com/76139600/172642360-e0221129-5381-4cd4-8d05-74d24e430702.mov


MiniDigger:
uuh, this looks cool

sulu5890:
I am generally happier with this now as it doesn't shift everything over extra. Another option I found is to use `box-shadow`, but I could not get that to work as well.

PR #83, opened 2022-10-02
Add a basic development guide

olijeffers0n:
Adds a basic project setup for gradle as well as an outline of the `plugin.yml`. Although spigot already has one of these pages, with the potential for specific paper keys with Owen's PR and a hardfork soon, I deemed it proper to have a page here as well.

PR #84, opened 2022-10-02
Add docs for sending plugin messaging

DeJayDev:
Saw an uptick in people asking about Plugin Messaging and figured it was kinda weird to send new devs to the Spigot Wiki.

Obviously my first time contributing to the docs here, but I've decided put it under the Velocity section of the docs but all the code samples use the Bukkit API. Mostly because we don't have a Paper How-To Section (pending #83?!).

I intended on keeping it vague, but didn't want to lose focus on the proxy so this didn't become a Paper guide. However, I'm confident in this as a starting point and would love some ideas/feedback especially as I learn to develop the Paper tone of voice. In the future I'd be open to writing more "how-to" style docs like this one.

### DIFF
--- a/docs/velocity/dev/how-to/plugin-messaging.md
+++ b/docs/velocity/dev/how-to/plugin-messaging.md
@@ -1,0 +1,69 @@
+---
+slug: /dev/plugin-messaging
+---
+
+# Plugin Messaging
+
+Plugin messaging is a way for Paper plugins to communicate with clients, and when your servers are behind a proxy, it will allow your Paper plugins to communicate with the proxy server.
+
+By default, your Velocity server will respond to the `bungeecord:main` channel unless you have disabled `bungee-plugin-message-channel` in the Velocity configuration. Let's take a look at how plugin messaging works, and more importantly - how your Velocity server is already prepared to handle it.
+
+## Sending Plugin Messages
+
+First, we're actually going take a look at your Java server. Your Java server plugin will need to register that it will be sending on any given plugin channel. You'll should to do this alongside your other event listener registrations.
+
+```java
+public final class PluginMessagingSample extends JavaPlugin {
+
+    @Override
+    public void onEnable() {
+        getServer().getMessenger().registerOutgoingPluginChannel(this, "bungeecord:main");
+    }
+
+}
+```
+
+Now that we're registered, we can send messages on the `bungeecord:main` channel. 
+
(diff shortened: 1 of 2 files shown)


Owen1212055:
I would like to see if you could perhaps include the built in bungee types (as we still have to go to the spigot forums for those) and lastly how to actually listen to and send plugin messages on the proxy as well. As currently it seems to only really describe how it is on the server, more in depth than the proxy id argue.

docs/velocity/dev/how-to/plugin-messaging.md
```diff
@@ -0,0 +1,69 @@
...
+
+# Plugin Messaging
+
+Plugin messaging is a way for Paper plugins to communicate with clients, and when your servers are behind a proxy, it will allow your Paper plugins to communicate with the proxy server.
+
+By default, your Velocity server will respond to the `bungeecord:main` channel unless you have disabled `bungee-plugin-message-channel` in the Velocity configuration. Let's take a look at how plugin messaging works, and more importantly - how your Velocity server is already prepared to handle it.
+
+## Sending Plugin Messages
+
+First, we're actually going take a look at your Java server. Your Java server plugin will need to register that it will be sending on any given plugin channel. You'll should to do this alongside your other event listener registrations.
```
MiniDigger, line 13: Java server? why not paper server? lol

e-im (changes requested):
This feels like it would work significantly better as two separate but intertwined guides— one in the Paper section focusing on how to send and receive on plugin channels more generally, possibly also mentioning their characteristics (they piggyback off of player connections, so won't work without a player online), and then one in the Velocity section talking about first the BungeeCord compatibility aspect as outlined here, as well as how Velocity plugins can send and receive messages on plugin channels.

For the Paper section, I wouldn't at all talk about `bungeecord:main` or proxy specifics, but instead link to the description of that in the Velocity guide.

I totally see why you put this in the Velocity section, but I don't think it makes sense without expansion. I'd also like to if at all possible keep the Velocity section free of Bukkit specific examples- as Velocity is compatible with and supports much more.

This is just an idea/my opinion, though, so feel free to do it or not if you disagree.

MrDienns:
Perhaps include how to register custom channels & custom messages (also on Velocity side).

lynxplay:
Superseded by 16f8edeb40177d7aa07b755a819819731f92c5fb.
Thank you for the work nonetheless <3

PR #87, opened 2022-10-05
Event API documentation

olijeffers0n:
Adds initial event API documentation. It outlines:
- Registering listeners
- HandlerLists
- Custom Events

It's possible I got some implementation details incorrect... :)

I also started something of a "folder" structure with this, lmk if your feel it should be different

### DIFF
--- a/docs/paper/dev/event-api/custom-events.md
+++ b/docs/paper/dev/event-api/custom-events.md
@@ -1,5 +1,5 @@
 ---
-slug: /dev/event-api/custom-events
+slug: /dev/custom-events
 ---
 
 # Custom Events
@@ -16,11 +16,11 @@ This list is used to call the listeners when the event is called.
 :::info `getHandlerList`
 
 Although it is not inherited from `Event`, you need to add a `getHandlerList()` method and return the `HandlerList` for your event.
+Both methods are required for your event to work.
 
 :::
 
 ```java title="PaperIsCoolEvent.java"
-
 public class PaperIsCoolEvent extends Event {
 
     private static final HandlerList HANDLER_LIST = new HandlerList();
@@ -40,11 +40,10 @@ Now that we have created our event, we can add some functionality to it.
 Perhaps this will contain a message that will be broadcast to the server when the event is called.
 
 ```java title="PaperIsCoolEvent.java"
-
 public class PaperIsCoolEvent extends Event {
 
     private static final HandlerList HANDLER_LIST = new HandlerList();
-    private final Component message;
+    private Component message;
 
     public PaperIsCoolEvent(Component message) {
         this.message = message;
@@ -62,6 +61,10 @@ public class PaperIsCoolEvent extends Event {
     public Component getMessage() {
         return this.message;
     }
+    
+    public void setMessage(Component message) {
+        this.message = message;
+    }
 }
 ```
 
@@ -70,7 +73,6 @@ public class PaperIsCoolEvent extends Event {
 Now that we have created our event, we can call it.
 
 ```java title="ExamplePlugin.java"
-    
 public class ExamplePlugin extends JavaPlugin {
 
     // ...
@@ -78,6 +80,9 @@ public class ExamplePlugin extends JavaPlugin {
     public void callCoolPaperEvent() {
         PaperIsCoolEvent coolEvent = new PaperIsCoolEvent(Component.text("Paper is cool!"))
         coolEvent.callEvent();
+        // Plugins could have changed the message from inside their listeners here. So we need to get the message again.
+        // This event structure allows for other plugins to change the message to their taste. 
+        // Like, for example, a plugin that adds a prefix to all messages.
         Bukkit.broadcast(coolEvent.getMessage());
     }
 }
@@ -88,11 +93,10 @@ public class ExamplePlugin extends JavaPlugin {
 If you want to allow your event to be cancelled, you can implement the `Cancellable` interface.
 
 ```java title="PaperIsCoolEvent.java"
-
 public class PaperIsCoolEvent extends Event implements Cancellable {
 
     private static final HandlerList HANDLER_LIST = new HandlerList();
-    private final Component message;
+    private Component message;
     private boolean cancelled;
 
     // ...
@@ -112,7 +116,6 @@ public class PaperIsCoolEvent extends Event implements Cancellable {
 Now, when the event is called, you can check if it is cancelled and act accordingly.
 
 ```java title="ExamplePlugin.java"
-
 public class ExamplePlugin extends JavaPlugin {
 
     // ...
--- a/docs/paper/dev/event-api/custom-events.md
+++ b/docs/paper/dev/event-api/custom-events.md
@@ -12,7 +12,13 @@ This will allow for other people to listen for your custom events and add functi
 To create a custom event, you need to create a class that extends `Event`. Each event has a `HandlerList` that contains all the listeners that are listening for that event.
 
 This list is used to call the listeners when the event is called.
-    
+
+:::info `getHandlerList`
+
+Although it is not inherited from `Event`, you need to add a `getHandlerList()` method and return the `HandlerList` for your event.
+
+:::
+
 ```java title="PaperIsCoolEvent.java"
 
 public class PaperIsCoolEvent extends Event {
--- a/docs/paper/dev/event-api/custom-events.md
+++ b/docs/paper/dev/event-api/custom-events.md
@@ -1,0 +1,122 @@
+---
+slug: /dev/event-api/custom-events
+---
+
+# Custom Events
+
+Creating custom events is a great way to add functionality to your plugin. 
+This will allow for other people to listen for your custom events and add functionality to your plugin.
+
+## Creating a custom event
+
+To create a custom event, you need to create a class that extends `Event`. Each event has a `HandlerList` that contains all the listeners that are listening for that event.
+
+This list is used to call the listeners when the event is called.
+    
+```java title="PaperIsCoolEvent.java"
+
+public class PaperIsCoolEvent extends Event {
+
+    private static final HandlerList HANDLER_LIST = new HandlerList();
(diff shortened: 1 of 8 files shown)


docs/paper/dev/event-api/event-listeners.md
```diff
@@ -0,0 +1,125 @@
...
+    @EventHandler
+    public void onPlayerJoin(PlayerJoinEvent event) {
+        event.setCancelled(true);
+    }
+}
+```
+
+The above example will cancel the event, meaning that the player will not be able to join the server.
+Once an event is cancelled, it will not call any other listeners for that event. 
+For this reason you can add `ignoreCancelled = true` to the `@EventHandler` annotation to ignore cancelled events.
```
kennytv, line 113: Should combine these "will not call any other listeners unless they..."

docs/paper/dev/event-api/event-listeners.md
```diff
@@ -0,0 +1,125 @@
...
+The `MONITOR` priority is used to monitor the event, but not change it.
+
+:::
+
+The order of the priorities is somewhat counter-intuitive. The **higher** the priority, the **later** the event is called. 
+This is because the priorities are called in reverse order as the later listener can override the result of the earlier listener.
+
+## Event cancellation
+
+Events can be cancelled, which means that the event will not happen.
```
kennytv, line 99:
```suggestion
Events can be cancelled, preventing the given action from being completed.
```

docs/paper/dev/event-api/event-listeners.md
```diff
@@ -0,0 +1,125 @@
+---
+slug: /dev/event-api/event-listeners
+---
+
+# Event Listeners
+
+Events are an efficient way to listen for specific actions that happen in the game. They can be called by the server, or by plugins. 
+These are called by the server when something happens, such as a player joining the server, or a block being broken.
```
kennytv, line 8: This is a bit clunky with "These are called by the server" and "They can be called by plugins" breaking the flow/repeating what was said right before

docs/paper/dev/event-api/custom-events.md
```diff
@@ -0,0 +1,122 @@
...
+# Custom Events
+
+Creating custom events is a great way to add functionality to your plugin. 
+This will allow for other people to listen for your custom events and add functionality to your plugin.
+
+## Creating a custom event
+
+To create a custom event, you need to create a class that extends `Event`. Each event has a `HandlerList` that contains all the listeners that are listening for that event.
+
+This list is used to call the listeners when the event is called.
```
kennytv, line 14: Should specifically call out that you need the static getHandlerList method in addition to the super class given method

docs/paper/dev/event-api/custom-events.md
```diff
@@ -0,0 +1,122 @@
...
+    
+```java title="PaperIsCoolEvent.java"
+
```
e-im, line 23: I'd get rid of the newlines at the beginning of each codeblock. It isn't what is done elsewhere and adds weird looking extra space. Let me know if there's a specific reason you have these though

docs/paper/dev/event-api/custom-events.md
```diff
@@ -0,0 +1,122 @@
...
+    private static final HandlerList HANDLER_LIST = new HandlerList();
+
+    public static HandlerList getHandlerList() {
+        return HANDLER_LIST;
+    }
+
+    @Override
+    public HandlerList getHandlers() {
+        return HANDLER_LIST;
+    }
```
e-im, line 35: Why is this duplicated? is what someone reading this might ask. I'd add something about that above saying you need both.

docs/paper/dev/event-api/custom-events.md
```diff
@@ -0,0 +1,122 @@
...
+```java title="ExamplePlugin.java"
+    
+public class ExamplePlugin extends JavaPlugin {
+
+    // ...
+
+    public void callCoolPaperEvent() {
+        PaperIsCoolEvent coolEvent = new PaperIsCoolEvent(Component.text("Paper is cool!"))
+        coolEvent.callEvent();
+        Bukkit.broadcast(coolEvent.getMessage());
```
e-im, line 81: Even though it is a bit obvious, for clarity, I'd add a comment here saying when other plugins will receive/modify message.

olijeffers0n:
Yeah... fixed my bad english

docs/paper/dev/event-api/event-listeners.md
```diff
@@ -0,0 +1,125 @@
...
+
+This will register your listener class and allow it to listen for events. 
+This is commonly done in the `onEnable()` method of your plugin so that it is registered when the server starts ticking.
+
+```java title="ExamplePlugin.java"
+public class ExamplePlugin extends JavaPlugin {
+
+    @Override
+    public void onEnable() {
+        Bukkit.getPluginManager().registerEvents(new ExampleListener(), this);
```
e-im, line 62: `getServer()`?

docs/paper/dev/event-api/event-listeners.md
```diff
@@ -0,0 +1,125 @@
...
+- `EventPriority.MONITOR`
+
+:::note
+
+The `MONITOR` priority is used to monitor the event, but not change it.
+
+:::
+
+The order of the priorities is somewhat counter-intuitive. The **higher** the priority, the **later** the event is called. 
+This is because the priorities are called in reverse order as the later listener can override the result of the earlier listener.
```
e-im, line 95: > This is because the priorities are called in reverse order

This is reiterating the line above, I'd take that part out and just describe why (second part of sentence): because higher priority means overriding what other plugins might have done -> called later.

I'd also move this to above the note about MONITOR.

docs/paper/dev/event-api/event-listeners.md
```diff
@@ -0,0 +1,125 @@
...
+- `EventPriority.LOWEST`
+- `EventPriority.LOW`
+- `EventPriority.NORMAL`
+- `EventPriority.HIGH`
+- `EventPriority.HIGHEST`
+- `EventPriority.MONITOR`
+
+:::note
+
+The `MONITOR` priority is used to monitor the event, but not change it.
```
e-im, line 90: I'd elaborate a bit here, say why you would want to use MONITOR to monitor the event. Because it is called last, so you get the result of possible modification/cancellation any other plugin performed.

docs/paper/dev/event-api/event-listeners.md
```diff
@@ -0,0 +1,125 @@
...
+
+:::
+
+The order of the priorities is somewhat counter-intuitive. The **higher** the priority, the **later** the event is called. 
+This is because the priorities are called in reverse order as the later listener can override the result of the earlier listener.
+
+## Event cancellation
+
+Events can be cancelled, preventing the given action from being completed.
```
e-im, line 99: Not all events can be

docs/paper/dev/event-api/event-listeners.md
```diff
@@ -0,0 +1,125 @@
...
+    @EventHandler
+    public void onPlayerJoin(PlayerJoinEvent event) {
+        event.setCancelled(true);
+    }
+}
+```
+
+The above example will cancel the event, meaning that the player will not be able to join the server.
+Once an event is cancelled, it will not call any other listeners for that event unless they add 
+`ignoreCancelled = true` to the `@EventHandler` annotation to ignore cancelled events.
```
e-im, line 113: > it will not call any other listeners

It **will** call every other listener, unless they add that. The default behavior is to continue down the list. I'd add a warning higher up about this behavior, to check cancellation/ignore if relevant as it trips up a lot of new people.

docs/paper/dev/getting-started/plugin-yml.md
```diff
@@ -1,5 +1,5 @@
 ---
-slug: /dev/plugin-yml
+slug: /dev/getting-started/plugin-yml
```
e-im, line 2: I'd drop this, to match the Velocity docs. If this is added though, also add a redirect in `vercel.json`

docs/paper/dev/getting-started/project-setup.md
```diff
@@ -1,5 +1,5 @@
 ---
-slug: /dev/project-setup
```
e-im, line 2: ditto

docs/paper/dev/event-api/event-listeners.md
olijeffers0n: Why is this better than using the `Bukkit` version?

kennytv: It is not strictly "better", but goes into OOP territory, also with the whole Bukkit class just being a relict and glorified getInstance() wrapper that simply duplicates all of the Server interface. using getServer() here simply nudges a little into the direction of good practices

docs/paper/dev/event-api/event-listeners.md
olijeffers0n: I believe I have understood this now, but do check please. Addressed everything else.

docs/paper/dev/event-api/custom-events.md
```diff
@@ -0,0 +1,131 @@
...
+
+## Creating a custom event
+
+To create a custom event, you need to create a class that extends `Event`. Each event has a `HandlerList` that contains all the listeners that are listening for that event.
+
+This list is used to call the listeners when the event is called.
+
+:::info `getHandlerList`
+
+Although it is not inherited from `Event`, you need to add a `getHandlerList()` method and return the `HandlerList` for your event.
```
e-im, line 18:
```suggestion
Although it is not inherited from `Event`, you need to add a static `getHandlerList()` method and return the `HandlerList` for your event.
```

e-im (approved):

docs/paper/dev/event-api/event-listeners.md
e-im: seems good

kennytv (approved):
Thanks!

PR #91, opened 2022-10-12
fix: improve grammar in "Paper System Properties"

kadenscott:
In the "Paper System Properties" page, under the first "INFO" block, there is some awkward grammar when referring to the presence of `.` when using Windows PowerShell. This PR clarifies that grammar.


e-im (commented):
"these will require wrapping in quotes" is also off. I'd say something closer to "you will need to wrap these in quotes"

olijeffers0n (approved):
I disagree, looks good as-is

Machine-Maker (approved):

PR #113, opened 2022-11-19
hide-itemmeta-with-visual-effects docs

NoahvdAa:
Docs for PaperMC/Paper#8561.


olijeffers0n:
> Controls whether item information that affects how an item is displayed (such as lodestone compass positions) should be sent to other players' clients. This information could be used by cheat clients to give another advantage to the cheating player. However, this may affect small details about how an item looks when held by another player.

Might be better? I feel like the first sentence was just way too long :)

NoahvdAa:
That sounds a lot better, thanks :)

olijeffers0n (approved):
Approved and waiting for Paper PR

Owen1212055 (approved):

PR #116, opened 2022-11-26
Add PDC Guide and Re-structure the developement Guides

olijeffers0n:
Adds an initial PDC development guide, a few things to mention though:

The drop-down it is in, currently called `Development Guides` to server as a container for general guides but might need renaming...

Also the content of the page. This is an initial draft so let me know with regards to spelling, grammar and also content

Thanks 😄


docs/paper/dev/guides/pdc.md
```diff
@@ -0,0 +1,132 @@
...
+- ##### World
+    - `World#getPersistentDataContainer()`
+- ##### Entity
+    - `Entity#getPersistentDataContainer()`
+- ##### TileState
+    - This is slightly more complicated, as you need to cast the block to something that extends `TileState`. 
+      This does not work for all blocks, only those that have a tile entity. F.E:
+      ```java
+        Block block = ...;
+        if (block instanceof Chest) {
```
Doc94, line 123: Consider the java version maybe can be good use `if (block instanceof Chest chest) {`? or like is is transversal maybe is better keep?

docs/paper/dev/guides/pdc.md
```diff
@@ -0,0 +1,132 @@
...
+NamespacedKey key = new NamespacedKey(pluginInstance, "example-key");
+
+ItemStack item = new ItemStack(Material.DIAMOND);
+// ItemMeta implements PersistentDataHolder so we can get the PDC from it
+ItemMeta meta = item.getItemMeta();
+item.getPersistentDataContainer().set(key, PersistentDataType.STRING, "I love Tacos!");
+item.setItemMeta(meta);
+```
+
+:::note NamespacedKeys
```
Doc94, line 41: not sure about the title.. mostly because in preview looks "strange" (for the uppercase)
image

docs/paper/dev/guides/pdc.md
```diff
@@ -0,0 +1,132 @@
...
+- [Chunk](#chunk)
+- [World](#world)
+- [Entity](#entity)
+- [TileState](#tilestate)
+- [ItemMeta](#itemmeta)
+- [Structure](#structure)
+
+## What is it used for?
+In the past, developers resorted to storing data within the NBT tags of an object. 
+This was a very hacky way of storing data, and was prone to breaking if the NBT tags were changed.
```
Lulu13022002, line 19: The last things is incorrect and it's always possible to break pdc if you change the nbt since pdc is stored in the nbttags. It's more about plugin conflict (with keys) than nbt change, the namespaced key resolve this issue

docs/paper/dev/guides/pdc.md
```diff
@@ -0,0 +1,132 @@
...
+ItemStack item = new ItemStack(Material.DIAMOND);
+// ItemMeta implements PersistentDataHolder so we can get the PDC from it
+ItemMeta meta = item.getItemMeta();
+item.getPersistentDataContainer().set(key, PersistentDataType.STRING, "I love Tacos!");
+item.setItemMeta(meta);
+```
+
+:::note NamespacedKeys
+
+It is considered good practice to store the `NamespacedKey`'s and reuse them.
```
Lulu13022002, line 43: This can confuse people since we talk about a store system so maybe just say that you should reuse the key but not store (some people can thinks about storing the key as value in the pdc) ? Also should be static

docs/paper/dev/guides/pdc.md
```diff
@@ -0,0 +1,132 @@
...
+    String value = container.get(key, PersistentDataType.STRING);
+    // Do something with the value
+    player.sendMessage(Compontent.text(value));
+}
+```
+
+## Data Types
+
+The PDC supports a wide range of data types, such as:
+`Byte`, `Byte` `Array`, `Double`, `Float`, `Integer`, `Integer Array`, `Long`, `Long Array`, `Short`, `String` as well as `Tag Containers`.
```
Lulu13022002, line 67: There some inconsistencies about the formatting. for example Byte Array has two labels but Integer array has only one label. Also consider maybe doing a list to improve readability like
- Byte, Byte Array
- Integer, Integer Array
- ...

Also you didn't explain how to nest a pdc inside another pdc using the PersistentDataAdapterContext#newPersistentDataContainer function

docs/paper/dev/guides/pdc.md
```diff
@@ -0,0 +1,132 @@
...
+        if (block instanceof Chest) {
+            Chest chest = (Chest) block;
+            chest.getPersistentDataContainer().set(key, PersistentDataType.STRING, "I love Tacos!");
+        }
+      ```
+- ##### Structure
+    - `Structure#getPersistentDataContainer()`
+- ##### ItemMeta
+    - `ItemMeta#getPersistentDataContainer()`
+    - `ItemStack#getItemMeta()#getPersistentDataContainer()`
```
Lulu13022002, line 132: This feels redundant with above

docs/paper/dev/guides/pdc.md
olijeffers0n: > It is considered good practice to reuse the `NamespacedKey`'s. Constructing them from a static context is ideal for this.

Is this better?

docs/paper/dev/guides/pdc.md
olijeffers0n: I've honestly never used this nested PDC structure. Is it something like: 

`container.set(key, PersistentDataType.TAG_CONTAINER, container.getAdapterContext().newPersistentDataContainer());`

docs/paper/dev/guides/pdc.md
Lulu13022002: Yes it's better

docs/paper/dev/guides/pdc.md
Lulu13022002: Yes but you don't need to show a full example but just how to create an empty pdc in order to nest bc it's a little bit hidden and non intuitive

olijeffers0n:
Addressed Comments :)

kennytv (approved):

docs/paper/dev/guides/pdc.md
```diff
@@ -0,0 +1,148 @@
...
+ItemStack item = new ItemStack(Material.DIAMOND);
+// ItemMeta implements PersistentDataHolder so we can get the PDC from it
+ItemMeta meta = item.getItemMeta();
+item.getPersistentDataContainer().set(key, PersistentDataType.STRING, "I love Tacos!");
+item.setItemMeta(meta);
+```
+
+:::note
+
+It is considered good practice to reuse the `NamespacedKey`'s. Constructing them from a static context is ideal for this.
```
NoahvdAa, line 44: I would clarify that you should reuse the objects, not the keys themselves. Otherwise (imo) it could be interpreted that you should reuse the same key for different information.

olijeffers0n: Better?

docs/paper/dev/guides/pdc.md
```diff
@@ -0,0 +1,148 @@
...
+## Getting Data
+To get data from the PDC, you need to know the `NamespacedKey` and the `PersistentDataType` of the data.
+
+```java
+// Create a NamespacedKey
+NamespacedKey key = new NamespacedKey(pluginInstance, "example-key");
+
+ItemStack item = ...; // Retrieve the item from before
+// Get the data from the PDC
+PeristentDataContainer container = item.getItemMeta().getPersistentDataContainer();
```
Lulu13022002, line 57: Small typo PeristentDataContainer instead of PersistentDataContainer

olijeffers0n: Thanks <3

NoahvdAa (approved):

Lulu13022002 (approved):

docs/paper/dev/guides/pdc.md
```diff
@@ -0,0 +1,148 @@
...
+
+- [Chunk](#chunk)
+- [World](#world)
+- [Entity](#entity)
+- [TileState](#tilestate)
+- [ItemMeta](#itemmeta)
+- [Structure](#structure)
+
+## What is it used for?
+In the past, developers resorted to storing data within the NBT tags of an object.
```
lynxplay, line 18: It might be nice to also include stuff like "checking the lore" or "checking the display name" here.
E.g. parsing integers from the lore.

Plugin made NBT tags were already the "better" solution lol :joy: most people we direct to the PDC come from a lore or display name usage.

docs/paper/dev/guides/pdc.md
```diff
@@ -0,0 +1,148 @@
...
+The second is a `PersistentDataContainer` which is the object you want to store the data on. The third is the data itself.
+
+```java
+// Create a NamespacedKey
+NamespacedKey key = new NamespacedKey(pluginInstance, "example-key");
+
+ItemStack item = new ItemStack(Material.DIAMOND);
+// ItemMeta implements PersistentDataHolder so we can get the PDC from it
+ItemMeta meta = item.getItemMeta();
+item.getPersistentDataContainer().set(key, PersistentDataType.STRING, "I love Tacos!");
```
lynxplay, line 38: I never knew ItemStack implemented PDH :)
```suggestion
meta.getPersistentDataContainer().set(key, PersistentDataType.STRING, "I love Tacos!");
```

docs/paper/dev/guides/pdc.md
```diff
@@ -0,0 +1,148 @@
...
+item.setItemMeta(meta);
+```
+
+:::note
+
+It is considered good practice to reuse the `NamespacedKey` objects. Constructing them from a static context is ideal for this.
```
lynxplay, line 44: This is a bit weird. Static context is rather troublesome for namespaced keys as they require your plugin instance which is not really available in a static context unless you do some fun static state.

Paper specifically un-deprecates https://jd.papermc.io/paper/1.19/org/bukkit/NamespacedKey.html#%3Cinit%3E(java.lang.String,java.lang.String), I am unsure if we want to promote its usage however.

docs/paper/dev/guides/pdc.md
```diff
@@ -0,0 +1,148 @@
...
+  // Get the existing container
+  PersistentDataContainer container = ...;
+  // Create a new container
+  PersistentDataContainer newContainer = container.getAdapterContext().newPersistentDataContainer();
+  ```
+  
+### Custom Data Types
+
+You can store a wide range of data in the PDC with the native adapters, however, if you need a more complex data type you can
+implement your own `PersistentDataType` and use that instead. Here is an example of how to do that for a UUID:
```
lynxplay, line 86: This should probably mention that a custom PDT's job is to boil a complex data structure down into one of the natively supported ones listed above.

The concept of the complex and primitive type and their respective constrains is otherwise unclear here.

docs/paper/dev/guides/pdc.md
electronicboy: personally, I've been moving towards using that constructor for various reasons

1) They're glorified strings, not instance derived magical things; purely making them instance based for the sake of injecting Plugin#gatName() as part of the glorified string is meh
2) stability, especially in terms of portability and renaming stuff; I.e. I can copy that key to another plugin and use it with 0 strife

I honestly think that that is the most appropriate constructor for creating keys in general, IMHO

lynxplay: Yea it is more of a "safety" for newer developers in my mind.
If you *know what you are doing*  that constructor is great. If you don't, better to stick to the plugin constructor than a `new NamespacedKey("minecraft", "my-key");`

docs/paper/dev/guides/pdc.md
electronicboy: It's sunday and not a wordsmith, but, this entire section is far from ideal, IMHO;
the nature of the PDC being better than NBT is really just about removing the need for abstractions/unstable reflection; it being "hacky" is really just weird, the only real place that supported arbitrary data in general was itemstacks, and that was purely on the foundation that the item tag stored arbitrary data anyways; until you get to the other things, entities, chunks, etc, which had 0 way to store arbitrary data on them in general

docs/paper/dev/guides/pdc.md
olijeffers0n: I have re-written this now

docs/paper/dev/guides/pdc.md
```diff
@@ -0,0 +1,156 @@
...
+// Create a NamespacedKey
+NamespacedKey key = new NamespacedKey(pluginInstance, "example-key");
+
+ItemStack item = ...; // Retrieve the item from before
+// Get the data from the PDC
+PersistentDataContainer container = item.getItemMeta().getPersistentDataContainer();
+if (container.has(key, PersistentDataType.STRING)) {
+    String value = container.get(key, PersistentDataType.STRING);
+    // Do something with the value
+    player.sendMessage(Compontent.text(value));
```
Lulu13022002, line 66: Small typo Compontent instead of Component

olijeffers0n: Thanks :)

e-im (commented):
I feel like this whole section (development) should match the layout of the velocity counterpart. The current Event API section is a guide too, so having a separate "guides" section for this doesn't make sense to me.

I think both of these should be moved into an `API` category, using titles more like "working with events" and "using persistent data containers". Or just "Event" and "PersistentDataContainer"

Might be a bit out of scope here, but if change is desired best to do it now before expanding the guides, such that fewer redirects are needed.

olijeffers0n:
That makes sense, would you like the event stuff to be kept as three separate guides under API/ or like keep them as a “sub category” of sorts? I will have a look tomorrow with regards to the redirects and will rename the PR accordingly 👍

olijeffers0n:
Looking at it, it doesnt need any redirects because I didn't add an "Event API" thing to the slug. I think that the Event stuff could be renamed / combined but i'm also don't mind it split up. Let me know if this isn't what you intended.

kennytv:
I'd open up a subcategory for events, otherwise people looking for how to unregister listeners will have no clue where to look as "Handler Lists" is meaningless for new devs and isn't referenced on the listeners page either

olijeffers0n:
> I'd open up a subcategory for events, otherwise people looking for how to unregister listeners will have no clue where to look as "Handler Lists" is meaningless for new devs and isn't referenced on the listeners page either

Sounds good, updated it!

kennytv:
Now just put event page above the pdc guide in the sidebar again, then it's good 😛

olijeffers0n:
😂

PR #119, opened 2022-12-17
[ci skip] Update badges in README

Doc94:
This PR handle a few things related to the badges in README.
- Removed the build badge because currently this rep not has active workflows.
- Added badge to discord
- Change the environment in the badge of deployment to point to the currently working production


olijeffers0n:
Maybe a line break between the title and the docs? I know this was how it was before so this is up to vote - otherwise looks good

Doc94:
> Maybe a line break between the title and the docs? I know this was how it was before so this is up to vote - otherwise looks good

not sure in the other repos the badges are in line with the title.

olijeffers0n (approved):
You’re right, they are

PR #120, opened 2022-12-17
Update filtering nbt docs

Machine-Maker:
For https://github.com/PaperMC/Paper/pull/6613

### DIFF
--- a/docs/paper/admin/reference/world-configuration.md
+++ b/docs/paper/admin/reference/world-configuration.md
@@ -444,9 +444,10 @@ Recommended configuration for both `engine-mode` `1` and `2` can be found in the
 
 #### filtered-entity-tag-nbt-paths
 
-- **default**: `[Pos, Motion]`
+- **default**: `[Pos, Motion, SleepingX, SleepingY, SleepingZ]`
 - **description**: A list of nbt paths that will be removed from the `EntityTag` tag on items which spawn entities.
-  The format of these strings follows the same format used to select nbt paths in vanilla commands. 
+  The format of these strings follows the same format used to select nbt paths in vanilla commands. If the spawning was
+  directly caused by a player and the player has the `minecraft.nbt.place` permission, the filter list will be ignored.
 - **note**: Some adventure maps may require this to be an empty list to function correctly, but we do 
   not recommend turning it off on a public server.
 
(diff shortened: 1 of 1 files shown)


docs/paper/admin/reference/world-configuration.md
```diff
@@ -440,7 +440,9 @@ Recommended configuration for both `engine-mode` `1` and `2` can be found in the
 
 - **default**: `[Pos, Motion]`
 - **description**: A list of nbt paths that will be removed from the `EntityTag` tag on items which spawn entities.
-  The format of these strings follows the same format used to select nbt paths in vanilla commands. 
+  The format of these strings follows the same format used to select nbt paths in vanilla commands. This will only
+  apply to place actions which do not have a player involved or for players who do not have the `minecraft.nbt.place`
+  permission. If a player with that permission places an entity, this filter list will be ignored.
```
MiniDigger, line 445: mmh, I feel like this could be better worded. can't we just say, "Players with the blah permission ignore this filter"?

Camotoy:
OK, so, even I (as someone who knows a bit of Minecraft) had to re-read this a couple times to understand it. But even so - maybe add why I, a server administrator, should care about this setting?

olijeffers0n:
I agree with Camotoy, I don't think the description does a very good job at all of describing what it does

docs/paper/admin/reference/world-configuration.md
```diff
@@ -448,8 +448,8 @@ Recommended configuration for both `engine-mode` `1` and `2` can be found in the
 - **description**: A list of nbt paths that will be removed from the `EntityTag` tag on items which spawn entities.
   The format of these strings follows the same format used to select nbt paths in vanilla commands. If the spawning was
   directly caused by a player and the player has the `minecraft.nbt.place` permission, the filter list will be ignored.
-  The defaults are set to prevent entities from spawning or moving to a place other than the location they are placed. If
-  `Pos` wasn't included, for example, a spawn egg could place an entity at any coords from any location.
+  The defaults are set to prevent entities from spawning or moving to a place other than the location they were placed. 
+  For example, if `Pos` wasn't included a spawn egg could place an entity at any location.
```
olijeffers0n, line 452: NitPicky, but you need a comma after included.

kennytv (changes requested):
nbt "tag(s)" instead of "path(s)" is the standard term used in the community

Machine-Maker:
> nbt "tag(s)" instead of "path(s)" is the standard term used in the community, including their names

Well the setting itself uses the word `paths` to avoid saying tag twice. Should the description still refer to them as tags?

kennytv:
A better name for the setting would have been "entity nbt tags", but yeah the description should just say tags to be easier to understand

Machine-Maker:
> entity nbt tags

would be easily mistaking for NBT tags on entities, which this is not. Its the `EntityTag` tag on ItemStacks

kennytv:
Oh righ, yeah. the desc should still be changed :p

kennytv (approved):

olijeffers0n (approved):

PR #122, opened 2022-12-21
Document paperclip.patchonly

Phoenix616:
This adds documentation for `paperclip.patchonly`

### DIFF
--- a/docs/paper/admin/reference/system-properties.md
+++ b/docs/paper/admin/reference/system-properties.md
@@ -194,3 +194,8 @@ It also remaps plugin CB calls to remove the version information.
 
 - **default**: `true`
 - **description**: Sets whether to parse comments in YAML files by default.
+
+#### paperclip.patchonly:
+
+- **default**: `false`
+- **description**: Sets whether the server should only patch the Vanilla server and download libraries.


MiniDigger:
mmmh, thats a paperclip property, not a paper one, I feel like they should be separate? but idk how to convey that without confusing people either, since the normal user its the same (by design)

I agree it should be documented here tho

NoahvdAa:
IMO it should just go on the Paper page since paperclip is only really used for Paper. Anyone running without paperclip will probably know what it is.

docs/paper/admin/reference/system-properties.md
```diff
@@ -194,3 +194,8 @@ It also remaps plugin CB calls to remove the version information.
 
 - **default**: `true`
 - **description**: Sets whether to parse comments in YAML files by default.
+
+#### paperclip.patchonly:
+
+- **default**: `false`
+- **description**: Sets whether the server should only patch the Vanilla server and download libraries.
```
NoahvdAa, line 201: I'd clarify that it quits after patching, this could be interpreted as that it only patches certain libs

MiniDigger:
> IMO it should just go on the Paper page since paperclip is only really used for Paper. Anyone running without paperclip will probably know what it is.

yeah, I was thinking some form of separator tho

docs/paper/admin/reference/system-properties.md
Phoenix616: Should be worded better now.

NoahvdAa (approved):

Machine-Maker:
I agree with Mini that there should be some sort of clarification that this will only work on a paperclip jar vs a paper server jar. I understand that's like 99.9% of all usecases, but the difference should be mentioned.

Phoenix616:
Added a mention for that now.

olijeffers0n (approved):

Machine-Maker (approved):

PR #123, opened 2022-12-29
Add missing values to nether configuration

davidjpfeiffer:
The recommended anti-xray config for the nether was missing the enabled and engine-mode values.

### DIFF
--- a/docs/paper/admin/how-to/anti-xray.md
+++ b/docs/paper/admin/how-to/anti-xray.md
@@ -141,7 +141,9 @@ anticheat:
     - ancient_debris
     - nether_gold_ore
     - nether_quartz_ore
+    lava-obscures: false
     max-block-height: 128
+    replacement-blocks: []
     update-radius: 2
     use-permission: false
 ```
--- a/docs/paper/admin/how-to/anti-xray.md
+++ b/docs/paper/admin/how-to/anti-xray.md
@@ -251,6 +251,7 @@ anticheat:
     - nether_gold_ore
     - nether_quartz_ore
     - polished_blackstone_bricks
+    lava-obscures: false
     max-block-height: 128
     replacement-blocks:
     - basalt
--- a/docs/paper/admin/how-to/anti-xray.md
+++ b/docs/paper/admin/how-to/anti-xray.md
@@ -142,6 +142,8 @@ anticheat:
     - nether_gold_ore
     - nether_quartz_ore
     max-block-height: 128
+    update-radius: 2
+    use-permission: false
 ```
 
 </details>
@@ -255,6 +257,8 @@ anticheat:
     - netherrack
     - soul_sand
     - soul_soil
+    update-radius: 2
+    use-permission: false
 ```
 
 </details>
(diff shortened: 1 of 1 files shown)


e-im:
This shouldn't be needed. It's inherited from the global config.

stonar96:
Yeah, originally the idea here was to show or teach people that they can setup a default config for all worlds and then do adjustments per world (inheritance). In the most cases this makes sense, so I have decided to intentionally omit everything that is inherited correctly. This has the disadvantage that it is assumed that the overworld (default) config is present (or that people don't just blindly copy parts without adjusting).

If we decide to change this, I'd also add all other omitted settings (e.g. `update-radius`, `use-permission`, ...) to the nether world config as well to be consistent. This would be the most foolproof solution, however, I'm afraid that the above described learning effect is lost and another disadvantage is that the config gets more decentralized (e.g. if one wants to temporarily disable Anti-Xray one has to edit multiple files and there's a chance to miss something). The end world config should of course stay as it is because adding settings if it's disabled would just lead to confusion.

davidjpfeiffer:
Okay, I just noticed that the instructions for The Overworld do specify that those values are for the `paper-world-defaults.yml` not the world-specific config.

For my server I need to configure anti-xray on a per-world basis. I searched Google for recommended anti-xray Paper config and clicked on this page from the docs. I copied The Overworld and The Nether configs into my world-specific config files and was confused why anti-xray wasn't working in The Nether, then realized that it was not enabled.

This might help people in similar situations, but up to you if you think this is helpful or not.

docs/paper/admin/how-to/anti-xray.md
```diff
@@ -135,11 +135,15 @@ Copy and paste into your `paper-world.yml` within your nether world folder. See
 ```yml title="world_nether/paper-world.yml"
 anticheat:
   anti-xray:
+    enabled: true
+    engine-mode: 1
     hidden-blocks:
     - ancient_debris
     - nether_gold_ore
     - nether_quartz_ore
     max-block-height: 128
```
stonar96, line 144:
```suggestion
    lava-obscures: false
    max-block-height: 128
```

Same for engine-mode: 2 below. Sorry, I can't do it via suggestion because it's not part of the diff.

docs/paper/admin/how-to/anti-xray.md
```diff
@@ -135,11 +135,15 @@ Copy and paste into your `paper-world.yml` within your nether world folder. See
...
+    update-radius: 2
```
stonar96, line 145:
```suggestion
    replacement-blocks: []
    update-radius: 2
```

stonar96 (changes requested):

davidjpfeiffer:
> Please note the additional note in the first review comment:
> 
> > Same for engine-mode: 2 below. Sorry, I can't do it via suggestion because it's not part of the diff.

Change for engine-mode: 2 was included in the commit

stonar96:
I can't see it. To be clear: `lava-obscures: false` is missing for engine-mode: 2 nether settings.

stonar96 (approved):

kennytv (approved):

PR #124, opened 2022-12-29
Change anti-xray world configuration headers

mja00:
This renamed the "Overworld Configuration" to default world since it's no longer the overworld config but a config for all worlds without an override.


stonar96:
Thanks! Either this PR or #123? (Or both?)

stonar96 (approved):

PR #127, opened 2023-01-03
Improve plugin.yml documentation

Phoenix616:
- Add `contributors` field
- Add default value of `load`
- Fix format of commands/permissions/children sections. Those need to be config sections, not map lists.
- Mention that permission default values that are not specified will default to the value of `default-permission` (which defaults to `op`)
- Add `default-permission` field
- Add `provides` field

Tbh. it's a bit questionable if this should be documented and maintained again on a separate page seeing as the JavaDocs [already provide a pretty good overview](https://jd.papermc.io/paper/1.19/org/bukkit/plugin/PluginDescriptionFile.html). Maybe that could be auto-generated at some point in the future?

### DIFF
--- a/docs/paper/dev/getting-started/plugin-yml.md
+++ b/docs/paper/dev/getting-started/plugin-yml.md
@@ -73,6 +73,12 @@ The author(s) of the plugin. This can be a single author or a list of authors.
 - `authors: [PaperMC, SpigotMC, Bukkit]`
 These will be shown in the plugin info commands.
 
+### contributors
+
+The contributors to the plugin that aren't the managing author(s).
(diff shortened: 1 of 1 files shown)


kennytv:
I think it's fine to have it laid out more with details and explanations on top of the quick (and ugly) jd pages; though it might make sense to also link to that table, like "If you only want a quick overview of the field names, you can also check the javadoc here" at the top

Machine-Maker (approved):
@Owen1212055 you think its still worth having this documented? We aren't dropping support for `plugin.yml`.

Owen1212055:
I see no reason to leave out docs like this, seems fine to me.

kashike (approved):

PR #128, opened 2023-01-07
Add tip about tabulators in the Anti-Xray guide and recommend backup

stonar96:


MiniDigger:
it feels kinda weird to specifically call out notepad++ here, is that still a state of the art editor people use? these also kinda more fit into a general "howto yaml" section and we could keep only a short hint and link to that here?

stonar96:
Thanks, I have removed the Notepad++ specific part and replaced it with a general note. Yeah, we can link to a section about YAML once we have one.

Machine-Maker (approved):

PR #132, opened 2023-01-19
Add InventoryHolder developer docs

Nacioszeczek:


docs/paper/dev/api/inventory-holder.md
```diff
@@ -0,0 +1,156 @@
+---
+slug: /dev/inventory-holder
+---
+
+# InventoryHolder
+
+Using an InventoryHolder is a great way of identifying your plugin's inventories in events.
```
Machine-Maker, line 7: I wouldn't call it "great". I think including those kinds of words doesn't fit in with a doc entry like this. Something like `InventoryHolder can be used to identify your plugins' inventories in events.`

docs/paper/dev/api/inventory-holder.md
```diff
@@ -0,0 +1,156 @@
...
+
+## Why use an InventoryHolder?
+
+InventoryHolder greatly simplifies the steps you need to do to make sure an inventory was created by your plugin.
```
Machine-Maker, line 11: no `greatly`.

docs/paper/dev/api/inventory-holder.md
```diff
@@ -0,0 +1,156 @@
...
+
+Using inventory names for identification is unreliable as other plugins can create inventories with names exactly as yours
+and with Components you need to make sure the name is exactly the same or serialize it to other formats.
```
Machine-Maker, line 14: This should be emphasized more. maybe some kind of colored box like note or info? Idk all the available ones.

docs/paper/dev/api/inventory-holder.md
```diff
@@ -0,0 +1,156 @@
...
+
+InventoryHolder has no such downsides and by using it you're guaranteed to have methods available to handle your inventory.
+
+## Creating an InventoryHolder
+
+InventoryHolder is an interface that we must implement.
+We can do this the following way, let's create a new class that will create our `Inventory` in the constructor.
```
Machine-Maker, line 21:
```suggestion
We can do this the following way: create a new class that will create our `Inventory` in the constructor.
```

docs/paper/dev/api/inventory-holder.md
```diff
@@ -0,0 +1,156 @@
...
+
+:::info
+
+The constructor takes your main plugin class as an argument in order to create the `Inventory`,
+if you wish you can use the static method `Bukkit.createInventory(InventoryHolder, int)` instead and remove the argument.
```
Machine-Maker, line 26:
```suggestion
If you wish, you can use the static method `Bukkit.createInventory(InventoryHolder, int)` instead and remove the argument.
```

docs/paper/dev/api/inventory-holder.md
```diff
@@ -0,0 +1,156 @@
...
+		return;
+	}
+
+	// Do what we need in the event.
+}
+```
+
+## Storing data on the InventoryHolder
+
+Imagine that we need to store some data for an inventory, if we're using InventoryHolders that
```
Machine-Maker, line 90:
```suggestion
Imagine that we need to store some data for an inventory, if we're using InventoryHolder that
```

docs/paper/dev/api/inventory-holder.md
```diff
@@ -0,0 +1,156 @@
...
+becomes extremely easy.
```
Machine-Maker, line 91: Remove "extremely easy". Rephrase this to sound less opinionated (I think that's the word to describe this). Something like `You can store extra information on an Inventory if you use an InventoryHolder.`

Machine-Maker (changes requested):

lynxplay (commented):
I wonder if we should mention that the InventoryHolder interface in general is also used by the server for things like "give me the chests that this inventory is from".

This is more of a "custom inventory holder/plugin owned inventory holder".
I only mention this because upstream had a rather long discussion and PR about removing the possibility of custom inventory holder a while back.

See also: https://hub.spigotmc.org/stash/projects/SPIGOT/repos/craftbukkit/pull-requests/482/overview

docs/paper/dev/api/inventory-holder.md
```diff
@@ -0,0 +1,159 @@
...
+Once we have the inventory open, we can listen to any inventory events we like and check if the `Inventory#getHolder()`
+returns an instance of our `MyInventory`.
+
+```java
+@EventHandler
+public void onInventoryClick(InventoryClickEvent event) {
+	Inventory inventory = event.getInventory();
+	// Check if the holder is our MyInventory,
+	// if yes, use instanceof pattern matching to store it in a variable immediately.
+	if(!(inventory.getHolder() instanceof MyInventory myInventory)) {
```
Lulu13022002, line 83: Can you put a space between `if` and `(` to keep the consistency of the doc

docs/paper/dev/api/inventory-holder.md
```diff
@@ -0,0 +1,159 @@
...
+
+}
+```
+
+Now, we can modify our listener to check if the player clicked the stone, and if so, add a click.
+
+```java
+@EventHandler
+public void onInventoryClick(InventoryClickEvent event) {
+	Inventory inventory = event.getInventory();
```
Lulu13022002, line 137: There's a little side effect using getInventory instead of getClickedInventory that means when the player will click to a stone in its own inventory (not in your custom inv), the counter will still increase.

docs/paper/dev/api/inventory-holder.md
Nacioszeczek: Do you mean just this `if` or all of them? This entire page doesn't have spaces between those

docs/paper/dev/api/inventory-holder.md
Nacioszeczek: This was of course not the intended purpose, but perhaps it would be better to either leave that to figure out for the reader themselves or further explain that in some other series about inventories similar to how `Event API` currently is

docs/paper/dev/api/inventory-holder.md
Lulu13022002: All of them, sorry if it wasn't clear

docs/paper/dev/api/inventory-holder.md
Lulu13022002: Hmm, i personally think that clicked inventory is more explicitly easy to understand that inventory (that currently always get the top inventory). But just want to make sure you were aware of that, i also think that a doc shouldn't lead to unexpected behavior that differ to what the doc said previously (in this case "the amount of times we clicked a stone **inside it**"). But i would be curious to see what others thinks

docs/paper/dev/api/inventory-holder.md
Nacioszeczek: Fixing next push

docs/paper/dev/api/inventory-holder.md
Nacioszeczek: Since there aren't any other opinions and I agree that we should keep to what was said committing with the `getClickedInventory()` now

Lulu13022002: There currently an NPE when you click outside of the window i think a (inventory == null || !(inventory.getHolder() instanceof MyInventory myInventory) should be added

Nacioszeczek: I highly doubt instanceof throws NPE on null

Lulu13022002: No it doesn't but the inventory var could be null in this case you call the getHolder on a null pointer

Nacioszeczek: Ah, got it

Nacioszeczek:
@lynxplay how would you like to resolve, something like rename the page to `Custom InventoryHolder` or add a section about using it this way?

Machine-Maker (approved):

lynxplay:
> @lynxplay how would you like to resolve, something like rename the page to `Custom InventoryHolder` or add a section about using it this way?

Yea I think "Custom Inventory Holder" is a better name. `Custom Inventories via InventoryHolder` might also work.
Generally just, avoiding confusion between the intended use and this "hacky" one.

docs/paper/dev/api/custom-inventory-holder.md
```diff
@@ -0,0 +1,162 @@
+---
+slug: /dev/custom-inventory-holder
+---
+
+# Custom InventoryHolder
+
+InventoryHolder can be used to identify your plugin's inventories in events.
```
lynxplay, line 7:
```suggestion
Custom InventoryHolders can be used to identify your plugin's inventories in events.
```

docs/paper/dev/api/custom-inventory-holder.md
```diff
@@ -0,0 +1,162 @@
...
+
+## Why use an InventoryHolder?
+
+InventoryHolder simplifies the steps you need to do to make sure an inventory was created by your plugin.
```
lynxplay, line 11:
```suggestion
Custom InventoryHolders simplify the steps you need to do to make sure an inventory was created by your plugin.
```

docs/paper/dev/api/custom-inventory-holder.md
```diff
@@ -0,0 +1,162 @@
...
+
+:::info
+
+Using inventory names for identification is unreliable as other plugins can create inventories with names exactly as yours
+and with Components you need to make sure the name is exactly the same or serialize it to other formats.
```
lynxplay, line 16:
```suggestion
and with components you need to make sure the name is exactly the same or serialize it to other formats.
```

docs/paper/dev/api/custom-inventory-holder.md
```diff
@@ -0,0 +1,162 @@
...
+
+InventoryHolder has no such downsides and by using it you're guaranteed to have methods available to handle your inventory.
```
lynxplay, line 18:
```suggestion
Custom InventoryHolders have no such downsides and by using them you're guaranteed to have methods available to handle your inventory.
```

docs/paper/dev/api/custom-inventory-holder.md
```diff
@@ -0,0 +1,162 @@
...
+
+:::
+
+## Creating an InventoryHolder
```
lynxplay, line 22:
```suggestion
## Creating a custom InventoryHolder
```

docs/paper/dev/api/custom-inventory-holder.md
```diff
@@ -0,0 +1,162 @@
...
+	if (!(inventory.getHolder() instanceof MyInventory myInventory)) {
+		// It's not our inventory, ignore it.
+		return;
+	}
+
+	// Do what we need in the event.
+}
+```
+
+## Storing data on the InventoryHolder
```
lynxplay, line 92:
```suggestion
## Storing data on the custom InventoryHolder
```

docs/paper/dev/api/custom-inventory-holder.md
```diff
@@ -0,0 +1,162 @@
...
+
+You can store extra data for your inventories on the InventoryHolder by adding fields and methods to your class.
```
lynxplay, line 94:
```suggestion
You can store extra data for your inventories on the custom InventoryHolder by adding fields and methods to your class.
```

lynxplay (dismissed):

NoahvdAa (approved):

PR #135, opened 2023-01-25
Fixed errors in Event-API document

MC-XiaoHei:
After actual measurement, in the new API (1.19.3-R0.1-SNAPSHOT), PlayerJoinEvent does not use the Cancellable interface described in the document. The code in the document is in. I think it should be changed and an alternative solution should be provided.

### DIFF
--- a/docs/paper/dev/api/event-api/event-listeners.md
+++ b/docs/paper/dev/api/event-api/event-listeners.md
@@ -111,7 +111,7 @@ public class ExampleListener implements Listener {
     }
 }
 ```
-**The BlockLockCheckEvent method and the PlayerLoginEvent method do not implement the Cancellable interface, but similar effects can still be achieved through the setResult (Result result) method.**
+**The BlockLockCheckEvent and PlayerLoginEvent types do not implement the Cancellable interface, but similar effects can still be achieved through the setResult (Result result) method.**
 :::warning
 
 It is important to consider that another plugin could have cancelled or changed the event before your plugin is called.
--- a/docs/paper/dev/api/event-api/event-listeners.md
+++ b/docs/paper/dev/api/event-api/event-listeners.md
@@ -111,7 +111,6 @@ public class ExampleListener implements Listener {
     }
 }
 ```
-**The BlockLockCheckEvent and PlayerLoginEvent types do not implement the Cancellable interface, but similar effects can still be achieved through the setResult (Result result) method.**
 :::warning
 
 It is important to consider that another plugin could have cancelled or changed the event before your plugin is called.
@@ -119,7 +118,7 @@ Always check the event before doing anything with it.
 
 :::
 
-The above example will cancel the event, meaning that the player will not be able to join the server.
+The above example will cancel the event, meaning that the Anvil will not cause damage.
 Once an event is cancelled, it will continue to call any other listeners for that event unless they add 
 `ignoreCancelled = true` to the `@EventHandler` annotation to ignore cancelled events.
 
@@ -127,8 +126,20 @@ Once an event is cancelled, it will continue to call any other listeners for tha
 public class ExampleListener implements Listener {
 
     @EventHandler(ignoreCancelled = true)
-    public void onPlayerJoin(PlayerJoinEvent event) {
+    public void onAnvilDamagedEvent(AnvilDamagedEvent event) {
         // ...
     }
 }
 ```
+## Set event result
+
+The results of some events can be modified, To change some behaviors. These events have method 'setResult' and internal enumeration class `Result`.
+    
+```java title="ExampleListener.java"
+public class ExampleListener implements Listener {
+
+    @EventHandler
+    public void onPlayerJoin(PlayerLoginEvent event) {
+        e.setResult(PlayerLoginEvent.Result.KICK_OTHER);//This means that the player cannot access the server for "other reasons"
+    }
+}


electronicboy:
PJE has never been cancellable, it's pretty much just a placeholder; if we're going to get semanticy over what is actually implemented by the class used in such examples, I'd much rather a dummy class name be used

docs/paper/dev/api/event-api/event-listeners.md
```diff
@@ -106,12 +106,12 @@ Some events can be cancelled, preventing the given action from being completed.
...
 
     @EventHandler
-    public void onPlayerJoin(PlayerJoinEvent event) {
+    public void onAnvilDamagedEvent(AnvilDamagedEvent event) {
         event.setCancelled(true);
     }
 }
 ```
-
+**The BlockLockCheckEvent method and the PlayerLoginEvent method do not implement the Cancellable interface, but similar effects can still be achieved through the setResult (Result result) method.**
```
lynxplay, line 114:
```suggestion
**The BlockLockCheckEvent and PlayerLoginEvent types do not implement the Cancellable interface, but similar effects can still be achieved through the setResult (Result result) method.**
```

electronicboy: I'm not sure such random misnomer notes should exist within the event listener docs, at least, not pointing to specific events; I'd much rather something like this exist but as a more generic nod towards the JDs

MC-XiaoHei: Thank you very much. I think your statement is more reasonable.

MC-XiaoHei: > 我不确定事件侦听器文档中是否应该存在这样随机的用词不当的注释，至少不指向特定事件；我宁愿存在这样的东西，但作为对 JD 的更通用的点头



> I'm not sure such random misnomer notes should exist within the event listener docs, at least, not pointing to specific events; I'd much rather something like this exist but as a more generic nod towards the JDs

I advocate this change because I only found that these two classes exist in this case. I think that if there is no special explanation, developers may not know how to implement similar functions (at least I do)

lynxplay: > I'm not sure such random misnomer notes should exist within the event listener docs, at least, not pointing to specific events; I'd much rather something like this exist but as a more generic nod towards the JDs

I mean, we don't really have a type for "result based events" like the login event etc.
For all I care, we could write proper docs in the Cancellable interface or on the specific events.
Don't think that makes it invalid to at least point out that, next to the cancellation type, some events types also accept a result. Tho this is really not a hill for me to die on 😓

lynxplay:
> PJE has never been cancellable, it's pretty much just a placeholder; if we're going to get semanticy over what is actually implemented by the class used in such examples, I'd much rather a dummy class name be used

At least on my opinion, an actual type and proper event that people, especially those new to development, can associate with an in game event (like the anvil damage event, or maybe even something easier like EntityDamageEvent) can be beneficial over some "CancellableEventType" dummy type.

docs/paper/dev/api/event-api/event-listeners.md
MC-XiaoHei: > > I'm not sure such random misnomer notes should exist within the event listener docs, at least, not pointing to specific events; I'd much rather something like this exist but as a more generic nod towards the JDs
> 
> I mean, we don't really have a type for "result based events" like the login event etc. For all I care, we could write proper docs in the Cancellable interface or on the specific events. Don't think that makes it invalid to at least point out that, next to the cancellation type, some events types also accept a result. Tho this is really not a hill for me to die on 😓

I think I understand. I will make some changes.

docs/paper/dev/api/event-api/event-listeners.md
```diff
@@ -106,29 +106,40 @@ Some events can be cancelled, preventing the given action from being completed.
...
-
 :::warning
 
 It is important to consider that another plugin could have cancelled or changed the event before your plugin is called.
 Always check the event before doing anything with it.
 
 :::
 
-The above example will cancel the event, meaning that the player will not be able to join the server.
+The above example will cancel the event, meaning that the Anvil will not cause damage.
```
DeJayDev, line 121:
```suggestion
The above example will cancel the event, meaning that the Anvil will not take damage.
```

docs/paper/dev/api/event-api/event-listeners.md
```diff
@@ -106,29 +106,40 @@ Some events can be cancelled, preventing the given action from being completed.
...
+## Set event result
+
+The results of some events can be modified, To change some behaviors. These events have method 'setResult' and internal enumeration class `Result`.
+    
+```java title="ExampleListener.java"
+public class ExampleListener implements Listener {
+
+    @EventHandler
+    public void onPlayerJoin(PlayerLoginEvent event) {
+        e.setResult(PlayerLoginEvent.Result.KICK_OTHER);//This means that the player cannot access the server for "other reasons"
```
DeJayDev, line 143:
```suggestion
        e.setResult(PlayerLoginEvent.Result.KICK_OTHER);
```

Comments a little ugly but that's personal opinion.

DeJayDev (commented):
AnvilDamagedEvent is when the anvil itself is damaged, not when a falling anvil damages a player.

PR #139, opened 2023-03-09
Fix dimensions of updated logo

Doc94:
This PR only change the width because this cause a few visual issues with the updated logo.
This is mostly a patch because in teory the logo need to be adapted maybe in dinamyc form?


kashike:
#140

PR #153, opened 2023-04-09
Add disabled commands to Folia's FAQ

mja00:
Folia disables a handful of commands and currently to see them you have to dig through the patch or code. This'll just list them all out in the FAQ for easier viewing.


NotMyFault (approved):
That is beneficial.

PR #154, opened 2023-04-11
Change config table of contents to include all settings

Malfrador:
Changes the maximum heading level for headings included in the table of content for both Paper config pages. This means that all config settings are now included in the sidebar on the right, in a layout somewhat similar to the config file itself.

The config pages are currently a bit unstructured and its hard to find the setting you are looking for and not all settings showing up in the sidebar isn't helping.

Downside to this is that it makes the table of content quite long. 

Before:
chrome_deDG767GQO
After:
chrome_AHIPkjg3Dk

### DIFF
--- a/docs/paper/admin/reference/global-configuration.md
+++ b/docs/paper/admin/reference/global-configuration.md
@@ -1,5 +1,6 @@
 ---
 slug: /reference/global-configuration
+toc_max_heading_level: 5
 keywords:
   - paper-global.yml
   - paper.yml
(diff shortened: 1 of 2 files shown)


mja00:
I messed around with the CSS a little bit. I tried moving the margin around on the list. Setting `--ifm-toc-padding-vertical` to `0.2rem` makes the empty space less noticeable. Could probably even move it tighter.
Microsoft Edge 2023-04-10 at 22 17 23@2x

electronicboy:
not fond of the notion of making the TOC stupidly large, the entire point of a TOC is that it makes it easier to navigate, not produces an unwieldy list

not hard against this, but, defo somewhat iffy

NotMyFault:
> not fond of the notion of making the TOC stupidly large

Same here, thanks for your contribution nonetheless.

PR #157, opened 2023-04-25
Add more dependencies to the Java install docs

clrxbl:
Even though most of these dependencies are shipped with minimal installations of Debian (and even the Ubuntu Docker image), there's been a few reports of them missing on Ubuntu.


NotMyFault (approved):
Thanks!

PR #160, opened 2023-05-03
Update MDX to v2

e-im:
Updates MDX (https://mdxjs.com/) which is used for markdown.
Provides better commonmark interop and resolves numerous markdown related issues.
Also faster build times, but only non-negligible in dev.

Breaking changes:
- Admonitions have a new format. Rather than `:::type Name` they are `:::type[Name]`. This is in line with the commonmark [generic directive syntax proposal](https://talk.commonmark.org/t/generic-directives-plugins-syntax/444).
- .mdx and .md files are treated differently. To improve commonmark interop in files that don't need MDX features, .md files are now treated as commonmark(ish) only. To use mdx features a .mdx file must be used
- HTML (`<!--`) comments aren't supporteed in MDX (and comments in commonmark are just generally not well defined, but HTML comments still work there). MDX uses JSX comments (`{/* ... */}`).


olijeffers0n (approved):

e-im:
Had marked as draft due to changes with prism-react-renderer, but I have verified that this is correct.

NotMyFault (approved):

PR #161, opened 2023-05-12
Folia overview.md: rephrase sentences

vadcx:
Both teleportation paragraphs were worded in a complicated manner. I hope I understood it right for the rewrite. Particularly I changed this construction (from/to):

> completed by forcing
the entity being teleported to be added to the entity chunk specified
by the target location

> completed by ensuring the addition
of the teleporting entity to the destination chunk specified
by teleporation

There's no "entity chunk" right? It's just a "chunk"

For simple teleportation I changed "forcing" to "ensuring", because the logic (as described) does not change the previous outcome, so it's less in the way of "forcing" and rather "ensuring" it happened. "Forcing" is still used for the Portal TP where it practically cancels, made more clear with "no teleportation"

PS: I had to reflow the Portal teleportation paragraph, the diff is misleading.

### DIFF
--- a/docs/folia/admin/reference/overview.md
+++ b/docs/folia/admin/reference/overview.md
@@ -239,13 +239,13 @@ differs by having the player in the world at the time of respawn.
 Portal teleport differs from simple teleportation as portalling does
 _not_ know the exact location of the teleport. Thus, the transform step
 does not update the entity position, but rather a new operation is inserted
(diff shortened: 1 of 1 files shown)


olijeffers0n (approved):
Looks generally good to me

NotMyFault (approved):

PR #162, opened 2023-05-18
API Roadmap

Owen1212055:


Machine-Maker (changes requested):
Would it be better to have 2 separate pages, one for plans and one for deprecations/removals?

docs/paper/dev/api/obsolete-api-migration/material.md
```diff
@@ -0,0 +1,72 @@
...
+
+#### Before
+```java
+new ItemStack(Material.AIR);
+```
+#### After
+```java
+ItemStack.of(ItemType.AIR);
+```
+*Using static constructors is encouraged over the use of the built-in constructor.
```
Machine-Maker, line 26: I think all built-in constructors are deprecated in upstream's PR now.

docs/paper/dev/api/roadmap.md
```diff
@@ -0,0 +1,93 @@
...
+
+In the future, ItemStack will be converted to an interface that allows developers to use an underlying 
+ItemStack directly instead of going through the API implementation.
+
+#### Precautions
+- Avoid using the ItemStack constructor, instead use the static constructors
+    - `ItemStack.of(ItemType.DIAMOND_PICKAXE);`
+      - The ``ItemStack`` constructor will be removed at some point.
+- Avoid directly extending the ``ItemStack`` class.
+  - Custom implementations of this class are not supported
```
Machine-Maker, line 29: This isn't possible without using the deprecated constructor as the ItemType constructor is private.

docs/paper/dev/api/roadmap.md
```diff
@@ -0,0 +1,93 @@
...
+Teleport an entity with the `RETAIN_PASSENGERS` flag, allowing its passengers to be transferred with the entity.
+```java
+entity.teleport(location, TeleportFlag.EntityState.RETAIN_PASSENGERS);
+```
+
+## Deprecation Policy
+
+:::warning
+
+It is highly recommended that you avoid using any APIs that are marked for deprecation.
```
Machine-Maker, line 61: I would change the wording here to be `marked as deprecated`. Marked "for deprecation" sounds like they will be deprecated in the future. I would also stress that anything scheduled for removal (include 2 examples, one with the deprecated annotation, one with the ApiStatus one) **WILL** be removed in the future, while just being marked as deprecated does not mean it won't be removed, but it isn't currently scheduled. Also talk about fair lead teim being given after something is scheduled for removal and actual removal. And actual removal will only happen on major mc version updates.

Machine-Maker (changes requested):
Include a link to the javadoc deprecation list in a prominent spot, otherwise I think this is good.

docs/paper/dev/api/roadmap.md
Owen1212055: Yes, but it's still possible and something we should note if people do this.

docs/paper/dev/api/obsolete-api-migration/material.md
Owen1212055: Yep, will remove.

Owen1212055:
Going to move migration guides to a separate PR.

PR #177, opened 2023-06-11
Add Boolean PDC Type

olijeffers0n:
Upstream recently added the Boolean Persistent Data Type, reflects this in the docs


lynxplay (commented):
I wonder if we should move this to its own section, as Boolean is a pure utility type based on Byte. So no one thinks they could create PDT that primitives to boolean.

olijeffers0n:
I can't think of something that would primitive to boolean, but it's a fair point. I could move boolean to the bottom and add a note / info message

olijeffers0n:
Added an info note to the type

kennytv (approved):

PR #180, opened 2023-06-17
Migration Guides

Owen1212055:
Best to wait so we have actual guides finished.

### DIFF
--- a/docs/paper/dev/api/obsolete-migration/adventure.md
+++ b/docs/paper/dev/api/obsolete-migration/adventure.md
@@ -1,0 +1,5 @@
+---
+slug: /dev/adventure
+---
+
+# Adventure Migration Guide
--- a/docs/paper/dev/api/obsolete-migration/material.md
+++ b/docs/paper/dev/api/obsolete-migration/material.md
@@ -1,0 +1,70 @@
+---
+slug: /dev/material
+---
+
+# Material Migration Guide
+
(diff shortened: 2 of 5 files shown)


docs/paper/dev/api/obsolete-migration/adventure.md
```diff
@@ -0,0 +1,5 @@
+---
+slug: /dev/adventure
```
Machine-Maker, line 2: Should be more specific with this slug. Like adding a `migration/` section.

docs/paper/dev/api/obsolete-migration/material.md
```diff
@@ -0,0 +1,70 @@
+---
+slug: /dev/material
```
Machine-Maker, line 2: Same as above

Machine-Maker (changes requested):

docs/paper/dev/api/obsolete-migration/material.md
```diff
@@ -0,0 +1,70 @@
...
+---
+
+# Material Migration Guide
+
+Material was originally built when every single item had a corresponding block.
```
me4502, line 7: I don't believe this line is true? There were numerous items that did not have a block equivalent at that point in time. stick, flint, coal, string, most of the non-cake food items, etc

PR #182, opened 2023-06-21
Add permission documentation for /velocity version

RalphORama:
I was trying to figure out how to disable `/velocity version` for some players and had to [delve into the source code](https://github.com/PaperMC/Velocity/blob/dev/3.0.0/proxy/src/main/java/com/velocitypowered/proxy/command/builtin/VelocityCommand.java#L258) to find the permission node. 

This change adds a note that the permission node `velocity.command.info` controls access to the `/velocity version` command.

### DIFF
--- a/docs/velocity/admin/reference/commands.md
+++ b/docs/velocity/admin/reference/commands.md
@@ -20,7 +20,7 @@ active on the proxy using `/velocity plugins`, including the name, authors, and 
 
 ### `/velocity version`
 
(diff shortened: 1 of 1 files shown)


olijeffers0n (approved):
Whilst your motives for creating this PR are questionable as the velocity version command shouldn't really be top of your list to hide from users, documenting a permission node is generally a good thing so I will approve

PR #187, opened 2023-06-25
Reading Stacktraces and debugging Plugins

### LINKED ISSUE #108: Debugging, reading stacktraces

olijeffers0n:
Closes #108 

Adds a new "misc" section to the dev docs including guides on reading stack traces and attaching remote IDE debuggers.

One thing I would like input on is any additional debug methods. Drop me a message on discord or on this PR


docs/paper/dev/misc/debugging.md
```diff
@@ -0,0 +1,76 @@
...
+### Randomly printing to the console
+
+One of the most common debugging techniques is to print to the console. This is likely something you've done before as its very simple. 
+It has a few downsides though. It can be hard to find the print statements in the console, and it can be hard to remove them all when you're done debugging.
+
+When debugging, you can use `System.out.println("");` to print to the console. It is recommended to use your plugin's logger instead though, 
+as it will be easier to know which plugin the log has come from. This can be done simply with:
+
+```java
+plugin.getLogger().warning("SuperDuperBad Thing has happened");
```
Machine-Maker, line 14: I think we just want to point to `Plugin#getComponentLogger` or `Plugin#getSLF4JLogger` instead of the java.util.logging.Logger

docs/paper/dev/misc/debugging.md
```diff
@@ -0,0 +1,76 @@
...
+You can do this by clicking the dropdown next to the run button and clicking `Edit Configurations...`:
+
+![](./assets/config_dropdown.png)
+
+Then, click the `+` button in the top left and select `Remote JVM Debug`. You can then name the configuration whatever you want, and click `Apply`:
+
+![](./assets/config_add.png)
+
+Finally, copy the command line arguments from the window, and paste these into your server's startup script. Once you have done this, you can click `OK`.
+Once your server is running, you can use the bug icon in the top right to connect your debugger to the server:
```
Machine-Maker, line 43: I would be more specific about where to put the arguments in the startup script. There's often confusion about if arguments go before or after the `-jar` part of the startup command.

docs/paper/dev/misc/debugging.md
```diff
@@ -0,0 +1,76 @@
...
+
+![](./assets/debugger_connect.png)
+
```
Machine-Maker, line 46: I feel like in addition to just remote debugging (maybe only exclusively), we should include direct debugging? I rarely use remote debugging, only in specific situations where its required. Usually now, I just use jmp's run-server gradle plugin and run `runServer` gradle task with debug which automatically starts the debugger.

docs/paper/dev/misc/reading-stacktraces.md
```diff
@@ -0,0 +1,51 @@
...
+
+Stacktraces are a useful tool for debugging your code. They show you the exact line of code that caused an error, and the 
+line of code that called that line of code, and so on. This is useful because it allows you to see the exact path of execution that led to the error.
+
+### Example:
+
+Here is an example of a stacktrace which has been caused due to a `NullPointerException`:
+
+```
+[15:20:42 ERROR]: Could not pass event PluginEnableEvent to TestPlugin v1.0
```
Machine-Maker, line 15: I would include a blurb talking about this line. While it isn't in every stacktrace, it feels common enough to say this means an exception happened while handling X event from Y plugin.

docs/paper/dev/misc/reading-stacktraces.md
```diff
@@ -0,0 +1,51 @@
+# Reading Stacktraces
```
Machine-Maker, line 1: Is it worth generalizing this, or dumbing it down a little and putting it not in the dev docs, but somewhere else? It's useful for non-devs to understand the very basics so they can maybe identify some causes if its plugin or something they can handle themselves.

docs/paper/dev/misc/reading-stacktraces.md
```diff
@@ -0,0 +1,51 @@
...
+        at net.minecraft.server.MinecraftServer.loadWorld0(MinecraftServer.java:636) ~[paper-1.20.1.jar:git-Paper-49]
+        at net.minecraft.server.MinecraftServer.loadLevel(MinecraftServer.java:435) ~[paper-1.20.1.jar:git-Paper-49]
+        at net.minecraft.server.dedicated.DedicatedServer.initServer(DedicatedServer.java:308) ~[paper-1.20.1.jar:git-Paper-49]
+        at net.minecraft.server.MinecraftServer.runServer(MinecraftServer.java:1101) ~[paper-1.20.1.jar:git-Paper-49]
+        at net.minecraft.server.MinecraftServer.lambda$spin$0(MinecraftServer.java:318) ~[paper-1.20.1.jar:git-Paper-49]
+        at java.lang.Thread.run(Thread.java:833) ~[?:?]
+```
+
+-   Firstly, we can see on the top line, the cause of the exception:
+    > `java.lang.NullPointerException: Cannot invoke "Object.toString()" because "null" is null`
```
Machine-Maker, line 39: > because "player" is null

Machine-Maker (changes requested):

docs/paper/dev/misc/reading-stacktraces.md
kennytv: "Reading Stacktraces/Plugin Errors"? I'd still include "stacktrace" in the title

docs/paper/dev/misc/debugging.md
kennytv: Most people won't be testing their plugin in a vacuum/inside rundev but some shared test server, so remote debugging is still the more important section imo. But adding something for direct debugging would still be nice

docs/paper/dev/misc/debugging.md
```diff
@@ -0,0 +1,77 @@
+# Debugging your plugin
+
+Debugging your plugin is vital to being able to fix bugs and issues in your plugin. This page will cover some of the most common debugging techniques.
+
+### Randomly printing to the console
+
+One of the most common debugging techniques is to print to the console. This is likely something you've done before as its very simple. 
+It has a few downsides though. It can be hard to find the print statements in the console, and it can be hard to remove them all when you're done debugging.
```
kennytv, line 8:
```suggestion
One of the most common debugging techniques is to print to the console. This is likely something you've done before, as it's very simple. 
This has a few downsides, though. It can be hard to find the print statements in the console, and it can be hard to remove them all when you're done debugging. Most notably, you have to recompile your plugin and restart the server to add or remove debugging.
```

docs/paper/dev/misc/debugging.md
```diff
@@ -0,0 +1,77 @@
+# Debugging your plugin
+
+Debugging your plugin is vital to being able to fix bugs and issues in your plugin. This page will cover some of the most common debugging techniques.
+
+### Randomly printing to the console
```
kennytv, line 5:
```suggestion
### Printing to the console
```
I would also add a warning here to make sure it's immediately obvious this is not the ideal solution

docs/paper/dev/misc/debugging.md
```diff
@@ -0,0 +1,77 @@
...
+To set up a debugger in IntelliJ, you need to create a new run configuration.
+You can do this by clicking the dropdown next to the run button and clicking `Edit Configurations...`:
+
+![](./assets/config_dropdown.png)
+
+Then, click the `+` button in the top left and select `Remote JVM Debug`. You can then name the configuration whatever you want, and click `Apply`:
+
+![](./assets/config_add.png)
+
+Finally, copy the command line arguments from the window, and paste these into your server's startup script.
```
kennytv, line 42: Not sure how, but I would highlight the "paste these into your server's startup script", maybe also include an example startup command

kennytv (changes requested):

docs/paper/dev/misc/reading-stacktraces.md
Machine-Maker: I wasn't referring to the name, but the location of this inside the `dev/` docs section.

docs/paper/dev/misc/debugging.md
olijeffers0n: Added a note about this at the bottom. I'd rather not add too much documentation here, as it will be somewhat redundant when there is the project README

kennytv (approved):

Machine-Maker (approved):

PR #192, opened 2023-06-30
Document added block update configuration

lynxplay:
The recently merged pull request PaperMC/Paper#9368 allows server owners to disable block updates from affecting note blocks or tripwires, providing an easy way to developers/mapmakers to use them for custom blocks.

This commit provides plain documentation for these two options on the paper documentation page.


olijeffers0n (approved):
I think these are really niche config nodes, and the description, although brief, is fine because the only people this matters for are people looking for it. So yeah, approved

Machine-Maker (approved):

PR #208, opened 2023-08-05
Add missing plugin messaging channel

olijeffers0n:
Adds `MessageRaw` with component example


Nacioszeczek:
Perhaps using a JSON component serializer would be way cleaner than the JSON itself?

MiniDigger:
yeah, really dont wanna encourage hard coding json strings in any way

olijeffers0n:
Yeah we discussed this last night. Updated it to use adventure

MiniDigger (approved):

docs/paper/dev/api/plugin-messaging.md
```diff
@@ -194,3 +195,29 @@ This is not a recommended way to ban players due to the fact that there may not
...
+public class MyPlugin extends JavaPlugin {
+    
+    @Override
+    public void onEnable() {
+        this.getServer().getMessenger().registerOutgoingPluginChannel(this, "BungeeCord");
+        
+        Player player = ...;
+        ByteArrayDataOutput out = ByteStreams.newDataOutput();
+        out.writeUTF("MessageRaw");
+        out.writeUTF("ALL");
```
NoahvdAa, line 214: The text above the snippet make it sound like this is only going to one player, when this is being broadcasted to all players on the proxy. This should be clarified to mention that this should be the name of the target player or ALL.

docs/paper/dev/api/plugin-messaging.md
```diff
@@ -194,3 +195,29 @@ This is not a recommended way to ban players due to the fact that there may not
 but it is an example of how this can be used.
 
 :::
+
+#### MessageRaw
+
+The `MessageRaw` message type is used to send a raw chat component to a player. This is useful for sending clickable messages to players,
+or messages with hover events.
```
NoahvdAa, line 202: Maybe also mention sending to players that aren't on the same backend server as an extra use case.

NoahvdAa (changes requested):

olijeffers0n:
Thanks, Addressed comments

NoahvdAa:
Now that I think of it, "This is useful for sending clickable messages to players," sounds weird considering we (and Spigot) have native support for chat components. I feel like this might be a leftover in the bungee docs from a time where this native integration didn't exist yet. Maybe we're better off removing this entire part? I don't really see the purpose for this channel anymore apart from the aforementioned option of using it to send messages to players on a different server.

olijeffers0n:
Good point. I think this is more just bad wording...  but yeah the fact that it is a clickable component is not really significant anymore.

NoahvdAa (approved):

PR #210, opened 2023-08-08
Note clarifying port ordering

JACTC:


kennytv:
Why?

JACTC:
Because if one server doesn't have the port in order, you wouldn't be able to connect to it, and because I haven't seen it mentioned anywhere in the documentation I thought it would help people troubleshoot faster over a simple issue that is port ordering.

But if you prefer to not mention it or mentioning it in another way, its totally fine by me.

kennytv:
The port definition doesn't matter outside of a range of reserved or otherwise maybe already used ports. Your first server could have the port 5543, the next 25599, and another 25566

PR #212, opened 2023-08-09
Automate the versioning of doc contents

olijeffers0n:
This grabs the versions of projects from `api.papermc.io`. There is still some logic required todo with when you are opening an old doc like 1.19 or 1.18. This requests every 5 minutes to keep them up-to-date when the page is still open.


olijeffers0n:
Ok, this is ready for review now. The basic logic behind what version to display to the user is:

- If the URL contains the version string, then we load the maximum version for that major. IE the dev guide will display `1.19.4` if we are using the 1.19 versioned docs. 

- On the other hand, if we are using the latest docs then we can just return the highest Paper version. 

### SoftwareVersionFetcher

In order to use this, there are some caveats. When it is placed inline within markdown text, that is simple. We can simply import the TypeScript class and call the method. This will require the file to be `mdx` rather than `md`:

```mdx
import SoftwareVersionFetcher from "@site/src/minecraft-versioning/SoftwareVersionFetcher";

The valid versions are 1.13 - { SoftwareVersionFetcher.getMajorPaperVersion() }.
```

The `SoftwareVersionFetcher` supports: 
- `getMajorPaperVersion` -> `1.19`
- `getMajorMinorPaperVersion` -> `1.19.4`
- `getMaxPaperVersion` -> `1.20.1`

### Code Blocks:

The majority of the usage of the version comes from within code blocks. This presented an issue as, for obvious reasons, it does not parse the code block contents. For this, code blocks must be wrapped in a `VersionFormattedCode` component. This does not require importing and it very simple to use. 

```
	```bash
	java -agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=*:5005 -jar paper-1.20.1.jar nogui
	``` 
```
Becomes:
```
	<VersionFormattedCode language={"bash"}>
	```
	java -agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=*:5005 -jar paper-%%_MAJ_MIN_MC_%%.jar nogui
	```
	</VersionFormattedCode>
```

They support two placeholders:
- `%%_MAJ_MIN_MC_%%` -> `1.20.1`
- `%%_MAJ_MC_%%` -> `1.20`

### API Fetching

I have been informed that the Paper API is cached by Cloudflare so this will not really cause an issue for you with regards to request frequency. The version is reloaded by the client every 5 minutes or if they navigate between versioned Paper documentation. 

There may be some logic improvements that can be made however I believe everything checks out. I look forward to everyone's comments :)

src/minecraft-versioning/SoftwareVersionFetcher.tsx
```diff
@@ -2,8 +2,10 @@ import axios from 'axios';
 
 class SoftwareVersionFetcher {
 
-    public paperVersion: string;
-    public maxPaperVersion: string;
+    private paperVersion: string;
+    private maxPaperVersion: string;
+    private currentURLVersion: string = null;
```
Machine-Maker, line 7:
```suggestion
    private currentURLVersion: string | null = null;
```

docs/paper/dev/api/event-api/event-listeners.mdx
```diff
@@ -42,7 +44,8 @@ public class ExampleListener implements Listener {
 :::note[Events]
 
 There is no list of events that can be listened to, however take a look 
-[here](https://jd.papermc.io/paper/1.20/org/bukkit/event/Event.html) to see all events that extend `Event`.
+<a href={"https://jd.papermc.io/paper/" + SoftwareVersionFetcher.getMajorPaperVersion() + "/org/bukkit/event/Event.html"}>here</a>
+to see all classes that extend `Event`.
```
Machine-Maker, line 48: I think it would probably be very handy to have a component to make linking to specific javadocs easier. Just to avoid the boilerplate of `<a href=...`. It could take a parameter for the path to whatever, and the inner content would be the text to show instead like the a tag.

src/components/VersionFormattedCode.tsx
```diff
@@ -0,0 +1,29 @@
...
+    code = code.replace(/%%_MAJ_MC_%%/g, SoftwareVersionFetcher.getMajorPaperVersion());
+
+    // Replace %%_MAJ_MIN_MC_%% with the minor Minecraft version
+    code = code.replace(/%%_MAJ_MIN_MC_%%/g, SoftwareVersionFetcher.getMajorMinorPaperVersion());
+
+    return (
+        inline ?
+            <code>{code}</code>
+            :
+            <CodeBlock language={language}>{code}</CodeBlock>
```
Machine-Maker, line 27: I think, even if we aren't using them now, we should add all the possible props to CodeBlock like `title`, and `showLineNumbers`

src/theme/MDXComponents.tsx
```diff
@@ -0,0 +1,10 @@
+import React from 'react';
+import MDXComponents from '@theme-original/MDXComponents';
+import VersionFormattedCode from '@site/src/components/VersionFormattedCode';
+
+export default {
+    // Re-use the default mapping
+    ...MDXComponents,
+    // Add a custom mapping
+    VersionFormattedCode,
+};
```
Machine-Maker, line 10: newline at end for this and all new files.

Machine-Maker (changes requested):

olijeffers0n:
There is some buggy behaviour when switching between versions. Other than that there are some syntax changes I will explain when ready for final review

the456gamer:
if you are on a sub page of the docs, and click on the version selector in the top right, it takes you back to the root page on the current version.
this feels like it might be confusing, especially as it tries to keep you on the same page when switching versions

olijeffers0n:
This isn’t behaviour I have changed, but I am happy to discuss it over on discord as I’m not quite sure what you mean 😄

olijeffers0n:
I have added all the behaviour of a native code block and changed the way that the versions work.

## Embedded Versions:

Before, we had:
```mdx
import SoftwareVersionFetcher from "@site/src/minecraft-versioning/SoftwareVersionFetcher";

The valid versions are 1.13 - { SoftwareVersionFetcher.getMajorPaperVersion() }.
```

However this only supports synchronous `getMajorPaperVersion` calls, so I have had to move this to a component:
```mdx
The valid versions are 1.13 - <SoftwareVersion versionType={"maj"}/>
```
`versionType` can take: `maj | min | maj-min`

## Javadoc links:

Before, we had a jank `a` tag that embedded the version into the JD URL, now you can use:
```mdx
<VersionedJavaDocLink target={"/org/bukkit/event/Event.html"}>here</VersionedJavaDocLink>
```
Which will create a hyperlink with the text `here` for the appropriate version

olijeffers0n:
Requires a rebase. Also, i seem to have ignored velocity. I assume we want that as well?

olijeffers0n:
It should be finalised now. Ready for review

docs/velocity/dev/getting-started/creating-your-first-plugin.mdx
```diff
@@ -95,11 +98,13 @@ dependencies {
     annotationProcessor("com.velocitypowered:velocity-api:3.2.0-SNAPSHOT")
```
kennytv, line 98: Missed this one?

docs/velocity/dev/getting-started/creating-your-first-plugin.mdx
```diff
@@ -112,6 +117,7 @@ dependencies {
     annotationProcessor 'com.velocitypowered:velocity-api:3.2.0-SNAPSHOT'
```
kennytv, line 117: And this?

kennytv (approved):

olijeffers0n:
Whoops, Yep addressed that

Nacioszeczek:
Could we also automate stuff like the jar name, e.g. [here](https://docs.papermc.io/paper/getting-started#running-the-server)?

olijeffers0n:
For all intents and purposes, this PR is now finished so a review would be appreciated 😄 - Kenny's review was a while ago now

src/components/SoftwareVersion.tsx
```diff
@@ -0,0 +1,30 @@
+import React, {useEffect, useState} from "react";
+import SoftwareVersionFetcher from "../minecraft-versioning/SoftwareVersionFetcher";
+
+export default function SoftwareVersion({ versionType, project = "paper" }) {
+    const [fetched, setFetched] = useState(null);
```
Machine-Maker, line 5:
```suggestion
    const [fetched, setFetched] = useState<string>();
```

My LSP complains about the typing being wrong here, and this seems more correct.

src/components/SoftwareVersion.tsx
```diff
@@ -0,0 +1,30 @@
+import React, {useEffect, useState} from "react";
+import SoftwareVersionFetcher from "../minecraft-versioning/SoftwareVersionFetcher";
+
+export default function SoftwareVersion({ versionType, project = "paper" }) {
```
Machine-Maker, line 4: I don't know how `typescript` we want to go, but ideally the props would have types. 

I don't know if there is a better way than just defining a local interface like
```typescript
interface SoftwareVersionProps
	versionType: "maj-min" | "maj" | "min",
	project?: "paper" | "whatever",
}

export default function SoftwareVersion({ versionType, project = "paper" }: SoftwareVersionProps) {
}
```

src/components/VersionFormattedCode.tsx
```diff
@@ -0,0 +1,46 @@
+import React, {useEffect, useRef, useState} from "react";
+import CodeBlock from "@docusaurus/theme-classic/lib/theme/CodeBlock";
+import SoftwareVersionFetcher from "../minecraft-versioning/SoftwareVersionFetcher";
+
+export default function VersionFormattedCode({ language = "", title = "", showLineNumbers = false, children }) {
+    const [formattedCode, setFormattedCode] = useState(null);
```
Machine-Maker, line 6: Same comments here as on the SoftwareVersion component.

src/components/VersionedJavaDocLink.tsx
```diff
@@ -0,0 +1,21 @@
...
+            let version = await SoftwareVersionFetcher.getMajorVersion(project);
+            if (project === "velocity") {
+                version = version[0] + ".0.0";
+            }
+
+            setHref("https://jd.papermc.io/" + project + "/" + version + target);
+        })();
+    }, [target]);
+
+    return (
```
Machine-Maker, line 18: Does this need to have a return early if href isn't set yet?

Machine-Maker (changes requested):

src/components/VersionedJavaDocLink.tsx
olijeffers0n: Not quite sure what you mean here, could you elaborate please?

olijeffers0n:
Config Descriptions are now versioned. This should be everything feature wise!

src/components/VersionedJavaDocLink.tsx
Machine-Maker: idk, just looking at the pattern here https://github.com/PaperMC/docs/pull/212/files#diff-15117eb1f365fc91bf29f3791c31fab4d45ec0af8b3238dbf474e0bcc800526cR25

olijeffers0n: My thought process here was that you could just return the `<a>` tag regardless and the href would just load in the background. maybe one with an empty href so that you don't end up with a duff link whilst it loads?

Machine-Maker: ya, that's fine then.

docs/versioned/paper/content/1.19/admin/reference/paper-plugins.md
```diff
@@ -6,7 +6,7 @@ slug: /reference/paper-plugins
 
 This documentation page serves to explain all the new semantics and possible confusions that Paper plugins may introduce.
 
-*As a developer, you can get more information [here](docs/paper/dev/getting-started/paper-plugins.md)*
+*As a developer, you can get more information [here](../../dev/getting-started/paper-plugins.md)*
```
Machine-Maker, line 9: avoid relative paths

docs/versioned/paper/content/1.19/admin/reference/paper-plugins.md
```diff
@@ -24,7 +24,7 @@ Paper plugins only support being loaded by Paper's Plugin Loader and may use new
 
 ### How do I add Paper plugins?
 
-Paper plugins are added the same as Bukkit plugins, therefore, you can follow [this guide](docs/paper/admin/getting-started/adding-plugins.md)
+Paper plugins are added the same as Bukkit plugins, therefore, you can follow [this guide](../../admin/getting-started/adding-plugins.md)
```
Machine-Maker, line 27: relative path

Machine-Maker (changes requested):

docs/versioned/paper/content/1.19/admin/reference/paper-plugins.md
Machine-Maker: apparently we have to use relative paths otherwise the docs versioning breaks

Machine-Maker (approved):

PR #215, opened 2023-08-12
Add docs for player reloot time

Machine-Maker:


olijeffers0n:
What’s the unit?

lynxplay:
> What’s the unit?

The duration, supports s,m,h,d.

olijeffers0n:
Should it maybe say that? From a quick look at paper it seems like only one unit is permitted rather than 2h3m30s which is how I would use this.

lynxplay:
Probably yea :+1:

olijeffers0n (approved):

PR #216, opened 2023-08-13
Add Config for mushroom spread

olijeffers0n:
For [Paper#9592](https://github.com/PaperMC/Paper/pull/9592)


Machine-Maker (changes requested):
Just blocking cause I think the structure of the options is going to change based on https://github.com/PaperMC/Paper/pull/9599#discussion_r1292826471

PR #227, opened 2023-08-25
EventHandlers return void, not null

A248:


Machine-Maker:
This is the EventExecutor interface being implemented in the lambda right? I think it also has 2 parameters, the listener and the event.

Nacioszeczek:
> This is the EventExecutor interface being implemented in the lambda right? I think it also has 2 parameters, the listener and the event.

The code from the docs does not compile.

Nacioszeczek (approved):

olijeffers0n (approved):

PR #231, opened 2023-08-29
Remove doubly nested A tags that trigger browser errors

olijeffers0n:


kennytv (approved):

MiniDigger:
while at it, maybe add a title attribute with the name of the config option to all links for better accessibility and seo
since the links with the anchors come after the actual text, browsers and crawlers dont know what the link refers to.

PR #232, opened 2023-08-29
add button that expands/collapses all config descriptions

granny:
This PR adds a button that lets you expand or collapse all the config descriptions at once. I've also taken the liberty of fixing some of the linter warnings I was getting, as well as some weird issues that I ran into while attempting to implement this feature.

I've also attached a `showDescriptions` parameter to the Config component that defaults to `false`. Setting it to `true` will expand all the descriptions initially. It might be useful to show all descriptions for smaller configs by default, since they don't take up as much screen space.

### DIFF
--- a/src/components/Config.tsx
+++ b/src/components/Config.tsx
@@ -206,7 +206,7 @@ export default function Config({ data, separator = ': ', showDescriptions = fals
     let ymlData = yaml.load(data);
     return (
         <div>
-            <pre className='codeblock'>
+            <pre className='config-button-parent'>
                 <button onClick={() => setShowAllExpanded(!showAllDescriptions)} className={`config-button button button--secondary`}>{showAllDescriptions ? "Collapse All" : "Expand All"}</button>
                 {renderYamlData(ymlData, '', true, separator, showAllDescriptions)}
             </pre>
--- a/src/components/Config.tsx
+++ b/src/components/Config.tsx
@@ -206,7 +206,7 @@ export default function Config({ data, separator = ': ', showDescriptions = fals
     let ymlData = yaml.load(data);
     return (
         <div>
-            <pre className='config-button-parent'>
+            <pre className='relative-container'>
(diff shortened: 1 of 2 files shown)


olijeffers0n (approved):

src/components/Config.tsx
```diff
@@ -159,44 +168,48 @@ const YamlTreeNode = ({ root, key, parentKey, value, separator }) => {
...
 };
 
-export default function Config({ data, separator = ': ' }) {
+export default function Config({ data, separator = ': ', showDescriptions = false}) {
+    const [showAllDescriptions, setShowAllExpanded] = useState(showDescriptions);
     let ymlData = yaml.load(data);
     return (
         <div>
-            <pre>{renderYamlData(ymlData, '', true, separator)}</pre>
+            <pre className='codeblock'>
```
MiniDigger, line 209: codeblock as a classname is a bit generic here, considering its only position relative, maybe configblock?

MiniDigger (approved):
only a nitpick

src/components/Config.tsx
granny: The sole purpose of the "position relative" class is to make it possible for a "position absolute" class to position correctly in whatever element it resides in. I originally renamed it to `config-button-parent` before changing my mind and naming it `relative-container` since it could be useful in other places in the future.

granny: Just realized I could also call it `config-container` :man_shrugging:

src/css/custom.css
```diff
@@ -181,6 +193,17 @@ html[data-theme="dark"] {
   width: fit-content;
 }
 
+.relative-container {
```
olijeffers0n, line 196: This class isn’t used?

granny: fixed.. i got too hyperfocused on the naming lol

kennytv (approved):

olijeffers0n (approved):

PR #234, opened 2023-08-30
Make config nodes accessible through keyboard shortcuts

granny:
This PR replaces the anchor element used for the config nodes with the button element. `clean-btn` and `button--link` are applied to render the button the same as an anchor element, and `config-node` is used for any leftover stylings required to keep visual parity.

https://github.com/PaperMC/docs/assets/43185817/62efe4ca-4f5b-4e59-a836-8655f46ad953

(keystroke visualizer: https://github.com/mulaRahul/keyviz)

Near the end of the video I notice a weird inconsistency with the config descriptions for the `hidden-blocks` and `replacement-blocks` showing incorrectly in production compared to my dev environment. Not sure what's up with that.


MiniDigger (approved):
lgtm

PR #238, opened 2023-08-31
Mention disabling server.props online mode in forwarding guide

Warriorrrr:
Not *too* sure on the wording, so suggestions welcomed. Disabling online mode in the server.properties is mentioned in the getting started guide at the very bottom, but worth mentioning on this page aswell

do we also want to mention this in the legacy forwarding part?


olijeffers0n:
Probably best to use the direct link? Its `/paper/reference/server-properties#online_mode`

Warriorrrr:
That's better yeah, thanks

olijeffers0n (approved):

olijeffers0n (approved):

PR #242, opened 2023-09-12
add compression-level description

Leguan16:
This adds the necessary description for the new `compression-level` config option in `config/paper-global.yml`

See https://github.com/PaperMC/Paper/pull/9711

### DIFF
--- a/docs/paper/admin/reference/configuration/global-configuration.mdx
+++ b/docs/paper/admin/reference/configuration/global-configuration.mdx
@@ -218,6 +218,9 @@ misc:
     chat-executor-max-size:
       default: '-1'
       description: 'The maximum number of threads to allow in the chat thread pool. The default value is -1, which means no maximum'
+  compression-level:
+    default: 'default'
+    description: 'A higher compression level means less data transmitted for the cost of more CPU time.'
 packet-limiter:
   all-packets:
     action:
--- a/docs/paper/admin/reference/configuration/global-configuration.mdx
+++ b/docs/paper/admin/reference/configuration/global-configuration.mdx
@@ -220,7 +220,10 @@ misc:
       description: 'The maximum number of threads to allow in the chat thread pool. The default value is -1, which means no maximum'
   compression-level:
     default: 'default'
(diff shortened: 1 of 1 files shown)


Nacioszeczek:
Waiting for the changes to be merged into Paper

olijeffers0n (changes requested):
Blocking for paper merge. Also, can you doc what the default value is please? In the description that is please ☺️

Leguan16:
well, i checked if default values in the other places (where IntOr is used) are present and they aren't so I did not include it here as well for consistency.

lynxplay:
Not that -1 would be of much use to the docs. The default is up to the deflate impl

Leguan16:
> Not that -1 would be of much use to the docs. The default is up to the deflate impl

does that mean that including the value is unnecessary or not?

lynxplay:
Can include the -1 being default, just you need to also explain that -1 means the compression will choose a sensible compression level.

Leguan16:
> Can include the -1 being default, just you need to also explain that -1 means the compression will choose a sensible compression level.

Should I also mention that Paper is using the velocity compressor and that the default is 6?

olijeffers0n:
Personally i would say yes, as it gives people something to use as a reference to their values as well.

Leguan16:
I hope this makes sense lol

olijeffers0n:
Yeah looks ok 👌

docs/paper/admin/reference/configuration/global-configuration.mdx
```diff
@@ -218,6 +218,12 @@ misc:
     chat-executor-max-size:
       default: '-1'
       description: 'The maximum number of threads to allow in the chat thread pool. The default value is -1, which means no maximum'
+  compression-level:
+    default: 'default'
+    description: 'A higher compression level means less data transmitted for the cost of more CPU time.
```
powercasgamer, line 223:
```suggestion
    description: 'A higher compression level means less data transmitted at the cost of more CPU time.
```

Leguan16:
changed

powercasgamer (approved):

lynxplay (approved):

PR #247, opened 2023-10-06
Latest information update

Lo-velo:
Velocity is not in active development. This is true.
https://github.com/PaperMC/Velocity/issues/1093


powercasgamer (changes requested):
Velocity is still being worked on.

PR #254, opened 2023-10-10
Latest Java LTS

Potothingi:
I recently heard that Paper recommends 21, the latest Java LTS.


lynxplay:
I honestly don't really mind suggesting/having docs to install 21.
Yea only 17 is required but moving the community forward sounds like a generally smart idea.
We'd have to note that down somewhere tho for sure and this 100% needs input from other team members.

kashike (approved):

PR #266, opened 2023-11-12
Rework title pages to use mdx for doc cards

olijeffers0n:
image


Machine-Maker (approved):
Looks like the velocity landing page also isn't using the cards, maybe it should as well to be consistent? Regardless, this looks good

granny:
^ I'd say the same with the Miscellaneous landing page.

olijeffers0n:
Velocity does this already within the Admin/Dev. Added Misc and also realised i named the branch velocity instead of waterfall 🤦

PR #267, opened 2023-11-18
Add client side entities page

Sytm:
A lot of times people appear in #paper-dev trying to get some packet magic working for basic client side entities, so I figured it would be nice to have some documentation around the already existing API.

The only problem with this is that `Entity.setVisibleByDefault` and `Player.showEntity` are still marked as experimental.


olijeffers0n (changes requested):
Thanks for your addition. The naming is slightly misleading because these are server-side entities, not client-side. There is also a lot of room for improvement on the implementation side. Maybe if you note that this is a fairly quick and dirty way of doing this it would be better

olijeffers0n:
Superseded, thank you for your work!

PR #270, opened 2023-12-08
Bump Docusaurus to 3.0.1

olijeffers0n:
Doesn't contain bug fixes that directly affect us, but best to keep up to date


e-im (commented):
Looks good

PR #272, opened 2023-12-18
separate config specs into other files

Machine-Maker:
I think its best if the specs for the various configs are loaded from a separate file instead of inline. Gives benefits when editing the yml directly in editors and prepares for a time from when the specs are read from the web instead of hardcoded.

```[tasklist]
- [x] server.properties
- [x] spigot.yml
- [x] bukkit.yml
- [x] paper-global.yml
- [x] paper-world-defaults.yml
```


olijeffers0n:
Looks like a good start, the only thing I was wondering was if we wanted to maybe place the YML files within a subdirectory like we do with assets? Other than that this should be good

MiniDigger:
> prepares for a time from when the specs are read from the web
maybe we should define them as a json schema then?

Machine-Maker:
> Looks like a good start, the only thing I was wondering was if we wanted to maybe place the YML files within a subdirectory like we do with assets?

Yeah, we could do that. Probably doesn't have to be in the `docs/` directory

olijeffers0n:
Is there anything making this not ready for review?

olijeffers0n (approved):

PR #281, opened 2024-01-14
Update docusaurus to 3.1.0 for better broken anchor detection

olijeffers0n:


olijeffers0n:
image
This happens when we have broken links now. Because of the configuration component we have to have it on warn, not throw unfortunately

kashike (approved):

PR #290, opened 2024-01-24
Document Adventure Audiences

olijeffers0n:


MiniDigger:
made add something like that all command senders are an audience and that server, world and team are forwarding audiences?

MiniDigger (approved):

PR #292, opened 2024-01-27
Document our enabling of sand duplication

olijeffers0n:


config-specs/paper/paper-global.yml
```diff
@@ -424,6 +424,13 @@ unsupported-settings:
     description: >-
       Whether the server should allow players to overstack items in the
       grindstone
+  allow-unsafe-end-portal-teleportation:
+    default: 'false'
+    description: >-
+      This setting allows for exploits related to end portal teleportation to be
+      possible. This includes sand duplication, and other exploits. This
```
MiniDigger, line 431: this reads weird, if you only listing one, maybe we should just write something like
```suggestion
      This setting allows for exploits related to end portal teleportation to be
      possible, for example sand duplication. This
```

MiniDigger:
I feel like we should note somewhere that unsupported settings might vanish at any time

olijeffers0n:
> I feel like we should note somewhere that unsupported settings might vanish at any time

Sure, not sure if this is the place though. Might require modifying the `Config.tsx` to allow a "note" on non-leaf nodes

PR #293, opened 2024-01-30
feat: ConfiguredProperty for documenting "magic" values consistently

olijeffers0n:
Need some input on where to store these values. Also, what other values need documenting.

### DIFF
--- a/src/components/ConfiguredProperty.tsx
+++ b/src/components/ConfiguredProperty.tsx
@@ -1,0 +1,30 @@
+import React from 'react';
+
+const CONFIGURED_PROPERTIES = {
+    MIN_VELOCITY_JAVA: "17",
+}
+
+const CONFIGURED_PROPERTY_NAMES = Object.keys(CONFIGURED_PROPERTIES);
+
+export function isConfiguredProperty(propertyKey: string): boolean {
+    return CONFIGURED_PROPERTY_NAMES.includes(propertyKey);
+}
+
+export function getConfiguredProperty(propertyKey: string): string {
+    return CONFIGURED_PROPERTIES[propertyKey];
+}
+
+export default function ConfiguredProperty({ propertyKey }) {
+
+    if (!isConfiguredProperty(propertyKey)) {
+        return null;
+    }
+
+    const value = getConfiguredProperty(propertyKey);
+
+    return (
+        <span className="configured-property">
+            {value}
+        </span>
+    );
+}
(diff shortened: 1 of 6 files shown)


Leguan16:
maybe store it in a separate JSON? Where to store it is a different question. Maybe config dir? a new dir for that seems overkill. Root dir is also a bit meh but easily findable. I'd just say config to be honest.

docs/paper/admin/getting-started/getting-started.mdx
```diff
@@ -13,12 +13,12 @@ With the release of Minecraft 1.18, Paper now requires **Java 17** to run. We re
 
 :::
 
-| Paper Version  | Recommended Java Version |
```
Leguan16, line 16: Missed a spot in the admonition above.

olijeffers0n: thanks

olijeffers0n:
Ready for final review

src/components/ConfiguredProperty.tsx
```diff
@@ -0,0 +1,20 @@
...
+
+export function isConfiguredProperty(propertyKey: string): boolean {
+    return CONFIGURED_PROPERTY_NAMES.includes(propertyKey);
+}
+
+export function getConfiguredProperty(propertyKey: string): string {
+    return configuredProperties[propertyKey];
+}
+
+export default function ConfiguredProperty({ propertyKey }) {
```
zlataovce, line 13: not that familiar with React components, but couldn't this also be typed?

```suggestion
export default function ConfiguredProperty({ propertyKey }: { propertyKey: string }): string | null {
```

src/components/ConfiguredProperty.tsx
```diff
@@ -0,0 +1,20 @@
+import configuredProperties from '@site/config-specs/properties/properties.json';
+
+const CONFIGURED_PROPERTY_NAMES = Object.keys(configuredProperties);
+
+export function isConfiguredProperty(propertyKey: string): boolean {
+    return CONFIGURED_PROPERTY_NAMES.includes(propertyKey);
+}
+
+export function getConfiguredProperty(propertyKey: string): string {
```
zlataovce, line 9:
```suggestion
export function getConfiguredProperty(propertyKey: string): string | undefined {
```

src/components/ConfiguredProperty.tsx
```diff
@@ -0,0 +1,20 @@
+import configuredProperties from '@site/config-specs/properties/properties.json';
+
+const CONFIGURED_PROPERTY_NAMES = Object.keys(configuredProperties);
+
+export function isConfiguredProperty(propertyKey: string): boolean {
```
zlataovce, line 5: this name seems a bit iffy, maybe something like `hasConfiguredProperty`?

PR #295, opened 2024-02-02
fix: reword how plugins work and move category

olijeffers0n:


docs/paper/dev/getting-started/how-do-plugins-work.md
```diff
@@ -80,15 +80,14 @@ give the player named Notch a diamond. Note here that the arguments are `["Notch
...
-`paper.command.help` permission.
+`paper.command`. This provides a nice way to group permissions together.
 
 ## Configuration
 
 Plugins can have configuration files. These files are used to store data that the plugin needs to run. For example, a
 plugin that adds a new block to the game might have a configuration file that stores the block's ID. Configuration files
-are stored in the `plugins` folder in the server directory. Configuration files are written in YAML. See 
-[here](/paper/dev/plugin-configurations) for more information.
+should be stored in the `plugins` folder in the server directory. The server offers a YAML configuration API that can be
```
zlataovce, line 89: worth to note that they should be in the plugin's respective data directory in `plugins` and not the root

MiniDigger:
I assume this changes the url? if so, we need a redirect

olijeffers0n:
It does not change the slug

Andre601:
I personally would've gone with this change for the Permission section to make it clear that a plugin is responsible for defining child-/sub-permissions to be part of a main one (Also adds note about the wildcard (`*`) permission and its use in permission plugins):
```markdown
Permissions are a way to control who can run commands and who can listen to events. Permissions 
are registered by plugins and can be checked by other plugins. Permissions can be granted to players and groups. 
Permissions can have a hierarchical nature, if defined so by the plugin in their `plugin.yml`. For example, a 
plugin can define `example.command.help` as a sub-permission of `example.command`. This means that if a player 
has the `example.command` permission, they will also have the `example.command.help` permission.

::: note

Permission plugins can allow the usage of wildcard permissions using the `*` character to grant any permission 
or sub-permission available. For example, granting `example.command.*` through a permission plugin with 
wildcard support will grant access to all permissions starting with `example.command.` itself.

It is **not** recommended to use wildcard permissions, especially `*` (All permissions), as it can be a huge 
security risk, as well as potentially causing unwanted side-effects to a player. Use with caution.

:::
```

olijeffers0n:
Thank you, I have used your comment

zlataovce (approved):

Andre601 (approved):

PR #299, opened 2024-02-03
fix: Adjust Aikar's flags for high memory

olijeffers0n:
closes #298

PR #300, opened 2024-02-04
Cleanup Aikar's flags

### LINKED ISSUE #298: Aikar's Flags don't match the Start Script Generator
There is a section in https://docs.papermc.io/paper/aikars-flags, that says if you has 12G+ memory, you should adjust flags a little.  

Script Generator completely skips that part and just pastes static flags.  

> изображение

Nacioszeczek:
Seems to me the page about Aikar's flags is a bit dated.
Cleaning it up a little and making it a consistent page,
also closes #298 and #299.

### DIFF
--- a/docs/paper/admin/how-to/aikars-flags.md
+++ b/docs/paper/admin/how-to/aikars-flags.md
@@ -71,17 +71,17 @@ up much space (5MB)
 
 ## Technical Explanation of the Flags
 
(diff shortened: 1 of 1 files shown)


olijeffers0n (commented):
Think on the getting started page or something we provide an example using 2G. Can you bump this to 4G as I believe this is the amount we recommend as a realistic value

olijeffers0n (approved):

PR #310, opened 2024-02-25
Add some notranslate hints

electronicboy:
I have no sane means to test this, but, this *should* tell site translators to not translate certain aspects of the configuration, if we want to go on with this, there's probably several other things we'd want to apply this to, i.e. any cross-refs to the configuration page, for ex


electronicboy:
image

olijeffers0n (approved):

PR #317, opened 2024-03-05
style: fix Prettier configuration and reformat

zlataovce:
If you run the `format` script through `yarn` right now, you'll find yourself with a whole bunch of changes, even breaking documentation in some places; this PR aims to fix that.

Seems like Prettier [does not support MDX v2 nor v3](https://github.com/prettier/prettier/issues/12209) yet, so I chose to ignore all Markdown and MDX files for the time being.


.prettierignore
```diff
@@ -1,10 +1,4 @@
-/.docusaurus
-/.vscode
-/.yarn
-/build
-/node_modules
```
olijeffers0n, line 5: Not sure why we are including node_modules?

zlataovce: Prettier automatically looks inside `.gitignore`, so it's redundant to ignore it there as well

olijeffers0n (approved):

PR #321, opened 2024-03-17
Add documentation for list persistent data types

lynxplay:
The PDC API recently was expanded to cover list types. This commit adds a simple example for developer that either
 a) Want to store a list of a custom pdt
 b) Want to store a list of strings or other api defined types

The commit also provides a quick example of retrieving the data as the implementation allows fully type-safe access to the data.

Also drops some unneeded spaces at EOL.

### DIFF
--- a/docs/paper/dev/api/pdc.md
+++ b/docs/paper/dev/api/pdc.md
@@ -22,7 +22,7 @@ In the past, developers resorted to a variety of methods to store custom data on
 - Lore and display names: Prone to collisions as well as slow to access.
 
 The benefit of the PDC is that it allows for a more reliable and performant way to store arbitrary data on objects.
-It also doesn't rely on accessing server internals, so it's less likely to break on future versions. It also removes the need to 
+It also doesn't rely on accessing server internals, so it's less likely to break on future versions. It also removes the need to
 manually track the data lifecycle, as, for example with an entity, the PDC will be saved when the entity unloads.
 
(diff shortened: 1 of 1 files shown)


zlataovce (approved):

docs/paper/dev/api/pdc.md
```diff
@@ -86,17 +86,34 @@ The PDC supports a wide range of data types, such as:
...
+  // a list data type wrapping the string data type.
+  container.set(
+    key,
+    PersistentDataType.LIST.listTypeFrom(PersistentDataType.STRING),
+    List.of("a", "list", "of", "strings")
+  );
+
+  // Storing a list of strings in a container by using the api
+  // provided pre-definitions of commonly used list types.
+  container.set(key, PersistentDataType.LIST.strings(), List.of("a", "list", "of", "strings"));
```
olijeffers0n, line 101: Is it worth listing the inbuild ones? I know you can see them from the javadocs so I’m not sure here.

olijeffers0n (approved):

docs/paper/dev/api/pdc.md
lynxplay: That's a lot to list and kinda just mirrors the existing list. I wouldn't add it but if others think that is benefitial, sure

olijeffers0n: Fairs, no issue.

PR #327, opened 2024-03-26
Added BungeeForge note to Velocity server-compatibility.mdx

p0t4t0sandwich:
Specifically under the Minecraft Forge (1.7.2-1.12.2) section, BungeeForge can be used in situations where SpongeForge isn't supported, compatible, or if the server admin feels their server doesn't need SpongeForge.

### DIFF
--- a/docs/velocity/admin/reference/server-compatibility.mdx
+++ b/docs/velocity/admin/reference/server-compatibility.mdx
@@ -78,7 +78,8 @@ or [SpongeForge](https://spongepowered.org/downloads/spongeforge) mod correspond
 Minecraft Forge for Minecraft 1.7.2-1.12.2 is fully compatible with Velocity, as we make special
 provisions to synchronize client state with each server. However, we **strongly** recommend the use
 of SpongeForge, as it allows you to use legacy BungeeCord player info forwarding and generally
-improves proxy support in general.
+improves proxy support in general. [BungeeForge](https://github.com/caunt/BungeeForge) can be used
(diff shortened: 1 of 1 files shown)


olijeffers0n:
BungeeForge is a very small project which appears to be abandoned for about 3 months, I’m hesitant to recommend it. Can you shed some light on why you are making this PR - has it helped you before?

p0t4t0sandwich:
Lots of projects in this space have stalled since the Forge/Neo split, but the current release works fine. There's quite a few people in the Ambassador Discord that use BungeeForge to support legacy forwarding on newer MC versions, myself included (I would rather use modern forwarding, but I support some older servers in the same network). I've also used it for a couple of old 1.12.2 modpacks and an old 1.7.10 modpack that I wanted to hook up to my network that didn't mesh well with SpongeForge, so I see it as a viable alternative.
I've tested BungeeForge quite thoroughly on quite a few versions/server platforms and it seems to work as expected across the board: [ProxyCompatibility.md](https://github.com/p0t4t0sandwich/MinecraftNotes/blob/main/ProxyCompatibility.md#forge) (still more testing to be done, but it's quite time intensive)

olijeffers0n:
I would like to thank you for explaining your reasoning here, and we are happy for you to recommend this within the community, however, we do not feel comfortable merging this into our documentation at this time. Thank you for your contribution!

p0t4t0sandwich:
Yeah for sure! Thanks for considering it!

PR #328, opened 2024-03-27
Fix wrong version reference in API call

Delath:
MINECRAFT_VERSION="1.20.4" is surely a copy/paste oversight, we need to obtain the build of the LATEST_VERSION and not of the MINECRAFT_VERSION, in the next steps we are rightfully referring only to the LATEST_VERSION.

### DIFF
--- a/docs/misc/downloads-api.mdx
+++ b/docs/misc/downloads-api.mdx
@@ -45,12 +45,11 @@ This will get the latest stable build for the given project and Minecraft versio
 #!/bin/bash
 
 PROJECT="paper"
-MINECRAFT_VERSION="%%_MAJ_MIN_MC_%%"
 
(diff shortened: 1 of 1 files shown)


zlataovce:
in my opinion, it should rather use `MINECRAFT_VERSION` to match the example's description and the latest version query should be extracted into another example

olijeffers0n:
Hi! Thank you for opening this PR, however this was not the intention we had behind this example. As @zlataovce said, this is technically correct as-is, however this latest version query should be a different example. If you would like to work on this, please open another PR! 😄

Delath:
> Hi! Thank you for opening this PR, however this was not the intention we had behind this example. As @zlataovce said, this is technically correct as-is, however this latest version query should be a different example. If you would like to work on this, please open another PR! 😄 

He didn't say that this is correct as-is, because it's not, however i think he is right and there should be two different examples, i may work on this when i have time.

PR #331, opened 2024-03-30
chore: remove inline styling & use clsx consistently

granny:
This PR removes all instances of inline styling. Ternary operator calls in `className` have been replaced with the clsx() helper method for improved readability.

1. Inline styles calling the margin[...] properties have been updated to use Docusaurus' margin-[...]--sm and margin-[...]--md classes for consistent margin sizes.
1. Inline styles have been merged into pre-existing classes wherever possible.

Only visual inconsistency should be the Icons that originally had 8 pixel margins and the slider markers in the script generator.


src/components/Config.tsx
```diff
@@ -163,16 +164,23 @@ const YamlNodeWithDescription = ({
 
   return (
     <div
-      style={{ paddingLeft: `${root ? 0 : INDENT_SIZE}px` }}
+      style={{ paddingLeft: `${root ? 0 : INDENT_SIZE}px` }} // TODO: dynamic class
```
zlataovce, line 167: considering `INDENT_SIZE` is a constant, this can be safely extracted into a conditional class

src/components/Config.tsx
```diff
@@ -246,14 +251,10 @@ const YamlTreeNode = ({
     <div
       key={name}
       className={`highlight-config-node`}
-      style={{ paddingLeft: `${root ? 0 : INDENT_SIZE}px` }}
+      style={{ paddingLeft: `${root ? 0 : INDENT_SIZE}px` }} // TODO: dynamic class
```
zlataovce, line 254: same here

src/components/Config.tsx
```diff
@@ -163,16 +164,22 @@ const YamlNodeWithDescription = ({
...
     >
-      <div className={`description_word_wrap`} style={{ marginBottom: showDescription ? 10 : 0 }}>
+      <div className={"description_word_wrap"}>
         <button
           onClick={() => {
             setShowDescription(!showDescription);
           }}
-          className={`config-node with-value${showDescription ? "-active" : ""} clean-btn button--link notranslate`}
+          className={clsx(
+            "config-node",
```
MiniDigger, line 176: any particular reason all these have been split? cant have have them all in one string and only hte active one in another? or at least group stuff that belongs together, like config-node with-value and clean-btn button--link, no need to have this use to many lines

granny: no specific reason; I've merged all the individual class strings into one

granny:
A `dropdown-item-selected` class has also been added to this PR

granny:
The internal styles added to the slider markers has been replaced with flexbox styling. It's now more accurate than how it was before.

Any leftover internal styling and TODOs have now been removed.

zlataovce (approved):

olijeffers0n (changes requested):
image
image

First is the PR, Second is Prod. The Config Pages have changed slightly with respect to margins and highlight colours not being active.

granny:
1. Selected config options should now properly highlight like before.
2. The margin indenting being off is technically on purpose since docusaurus' built-in `margin-bottom--sm` class is not 10px, but something closer to 8px. With `margin-bottom--xs` being closer to 4px and `margin-bottom--md` being closer to 16px.

olijeffers0n:
Looks generally good, I’ll review properly soon

PR #335, opened 2024-03-31
Improve Start Script Generation for Windows

notTamion:
adds an @echo off in front of all windows scripts and adds a pause statement after if auto restart is disabled. This is because people on windows tend to double click the start.bat instead of opening a terminal and running the script from there, which causes the terminal to close when the server stops without giving the user the option to look at e.g. crash logs, i have noticed a lot of people on the discord server just saying "my server stops without any crash log" because of this exact reason.

### DIFF
--- a/src/components/StartScriptGenerator.tsx
+++ b/src/components/StartScriptGenerator.tsx
@@ -72,7 +72,7 @@ const generateStartCommand = (
 ) => {
   setTimeout(resizeOutput, 0);
   let content = "";
-  const command = `java -Xmx${memory * 1024}M -Xms${memory * 1024}M ${selectedFlag.value}${selectedFlag === FLAGS.NONE ? "" : " "}-jar ${filename === "" ? "server.jar" : filename} ${guiEnabled || selectedFlag === FLAGS.VELOCITY ? "" : "--nogui"}`;
+  const command = `java -Xmx${memory * 1024}M -Xms${memory * 1024}M ${selectedFlag.value}${selectedFlag === FLAGS.NONE ? "" : " "}-jar ${filename === "" ? "server.jar" : filename} ${guiEnabled || selectedFlag === FLAGS.VELOCITY ? "" : "nogui"}`;
 
   if (autoRestartEnabled)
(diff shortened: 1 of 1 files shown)


DerEchtePilz:
As a suggestion/question, would it make sense to put the pause statement behind a toggle that by default is turned on?

notTamion:
i didn't look into the component too much but as it looks to me that would require moving stuff around a lot so we can have different toggles for windows and linux. since i am very sure 90+% windows folks are gonna double click it i don't see much reason to have a toggle here.

DerEchtePilz:
I guess that's fair. I don't use the script generator anyway, and also didn't expect that much more work for a toggle so you have my approval (not that you need it, anyway)

zlataovce (approved):

src/components/StartScriptGenerator.tsx
```diff
@@ -79,9 +79,10 @@ const generateStartCommand = (
       "%%CONTENT%%",
       command
     );
-  else content = command;
+  else content = platform === "windows" ? command + "\n\npause" : command;
 
   if (platform === "linux") content = "#!/bin/bash\n\n" + content;
+  if (platform === "windows") content = "@echo off\n\n" + content;
```
olijeffers0n, line 85:
```suggestion
  if (platform === "linux") {
   content = "#!/bin/bash\n\n" + content;
  } else {
    content = "@echo off\n\n" + content;
  }
```

I would rather do this like this, or with a ternary. Up to you :)

olijeffers0n (changes requested):

notTamion:
i know this is technically completely out of scope for this pr (but opening a pr for this small thing is eeh) but shouldn't we have the gen add "nogui" instead of "--nogui" that way it will also work on older versions

olijeffers0n:
> i know this is technically completely out of scope for this pr (but opening a pr for this small thing is eeh) but shouldn't we have the gen add "nogui" instead of "--nogui" that way it doesn't crash on older versions

Yes, we should. Feel free to add that to this PR if you would like, and I will rename the commit on merge

olijeffers0n (approved):

PR #347, opened 2024-04-08
Adds a commit hash to the "Last Updated" text at the bottom of the page

granny:
TODO:

- [x] turn `src/util/gitUtils.ts` into a yarn patch
  - split the shell execution out into it's own method to make it easier to recreate `getFileCommitHash`
- [ ] move `parseFrontMatter` function elsewhere?
- [x] better naming for `LastCommitIn` component and it's translation key/values or omit translation component
- [x] fix up namings of everything
- [x] fix up TODOs
- [x] clean up stuff left over from debugging
- [x] turn `author` into a hyperlink
- [x] turn `hash` into a hyperlink
- [x] replace octokit with a simple axios http request


package.json
```diff
@@ -69,6 +70,8 @@
   },
   "packageManager": "yarn@4.1.1",
   "resolutions": {
-    "@docusaurus/plugin-content-docs@3.2.1": "patch:@docusaurus/plugin-content-docs@npm%3A3.2.1#./.yarn/patches/@docusaurus-plugin-content-docs-npm-3.2.1-7bedb7ec1c.patch"
+    "@docusaurus/plugin-content-docs@3.2.1": "patch:@docusaurus/plugin-content-docs@npm%3A3.2.1#./.yarn/patches/@docusaurus-plugin-content-docs-npm-3.2.1-7bedb7ec1c.patch",
+    "@docusaurus/utils@3.2.1": "patch:@docusaurus/utils@npm%3A3.2.1#./.yarn/patches/@docusaurus-utils-npm-3.2.1-cfade3be35.patch",
+    "@docusaurus/utils@3.1.0": "patch:@docusaurus/utils@npm%3A3.2.1#./.yarn/patches/@docusaurus-utils-npm-3.2.1-cfade3be35.patch"
```
olijeffers0n, line 75: What's going on here? Is there transitively 2 versions of utils?

src/util/authorUtils.ts
```diff
@@ -0,0 +1,66 @@
...
+      console.error(error.response?.data ?? error.response);
+    } else {
+      // silent
+      console.error(error);
+    }
+  }
+}
+
+export async function cacheAuthorData() {
+  // TODO: dirty hack to make sure it's only run before `parseFrontMatter` call
```
olijeffers0n, line 51: Is this the first call? If so a simple boolean check would suffice here

granny: a simple check would not work - each time the config is called it ends up re-running the method. (incrementing a variable every time it ran would always output `1`).

The config gets called and loaded 3 times separately - at the very beginning of the build, right before the `parseFrontMatter` method is called, and sometime near the very end of the build.

The best solution I can think of is checking a variable used by docusaurus that's in scope when the method is called.

docusaurus.config.ts
```diff
@@ -80,6 +83,28 @@ const config: Config = {
...
     },
     format: "detect",
+    parseFrontMatter: async (params) => {
+      const result = await params.defaultParseFrontMatter(params);
+      let author: AuthorData = {
+        ...AUTHOR_FALLBACK,
+      };
+      if (process.env.NODE_ENV !== "development") {
+        const { commit } = await getFileCommitHash(params.filePath);
+        let username = commitCache.get(commit);
```
MiniDigger, line 93:
```suggestion
        const username = commitCache.get(commit);
```

olijeffers0n (approved):

PR #353, opened 2024-04-20
Document find-structures-outside-world-border

notTamion:
PaperMC/Paper#10437


Machine-Maker (changes requested):
This description isn't sufficient to describe all the functionality this affects. It affects more than just treasure maps or ender eye items, it affects any logic that searches for structures in the world, like the `/locate` command or plugins doing structure searches.

Machine-Maker (approved):

PR #361, opened 2024-04-23
Document various 1.20.5 config changes

Leguan16:
For now, I added the 2 new `server.properties` entries and `plugin.yml` now accepts patches in their `api-version` field. Let me know if i missed something.

Blocked until 1.20.5 ready.

### DIFF
--- a/config-specs/paper/server-properties.yml
+++ b/config-specs/paper/server-properties.yml
@@ -1,3 +1,6 @@
+accept-transfers:
+  default: "false"
+  description: "Whether this server accepts transfers from other servers using the transfer command/packet. If this is set to false, the server will disconnect the client."
 allow-flight:
   default: "false"
   description: >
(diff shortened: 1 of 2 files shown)


config-specs/paper/server-properties.yml
```diff
@@ -1,3 +1,6 @@
+accept-transfers:
+  default: "false"
+  description: "Whether this server accepts transfers from other servers using the transfer command/packet. If this is set to false the server will disconnect the client"
```
zlataovce, line 3:
```suggestion
  description: "Whether this server accepts transfers from other servers using the transfer command/packet. If this is set to false, the server will disconnect the client."
```

config-specs/paper/server-properties.yml
```diff
@@ -167,6 +170,11 @@ rcon.password:
 rcon.port:
   default: "25575"
   description: "The port to start the rcon server on."
+region-file-compression:
+  default: "deflate"
+  description: >
+    Specifies the compression type used to compress region files. Possible values are: "deflate", "lz4" and "none".
+    If set to "none" region files will take up significantly more disk space but might make sense together with filesystem level compression.
```
zlataovce, line 177:
```suggestion
    If set to "none", region files will take up significantly more disk space, but might make sense together with filesystem-level compression.
```

Leguan16: would this make more sense?
```suggestion
    If set to "none", region files will take up significantly more disk space, but it might make sense together with filesystem-level compression.
```

zlataovce: 👍

olijeffers0n (approved):

PR #365, opened 2024-04-28
Update toolchain version to Java 21 in Paper project setup

NoahvdAa:


olijeffers0n (approved):

olijeffers0n:
In the future, we should allow for properties in code blocks, however this works for now. Thanks

PR #375, opened 2024-05-05
feat: add support for entity argument to command converter, use float…

MiniDigger:
…ing-ui for select and new autocomplete


olijeffers0n:
I find it counter intuitive that they change locations, rather than just adding a new box. need other's input here but personally i'm not a fan

kennytv:
I'd maybe have the entity type box appear in the middle, so that the mode doesn't move

olijeffers0n (approved):

PR #376, opened 2024-05-05
feat: add mode to convert functions

MiniDigger:


kennytv:
There should also be an additional warning that macros can't be handled and that multi-line commands will be absorbed into a single line

Strokkur424:
Do we still care about this? I feel like 1.20.5 has been long enough already. We could probably also remove the item converter as a whole, but I think there is no real reason to do that.

PR #408, opened 2024-06-04
Update api-version usage with minor version support.

rainbowdashlabs:
Spigot changed the api version in 1.20.5 and allows minor versions from there on. Since paper adapted this change the docs should reflect this, which they in fact already do. So the documentation is not coherent with the example already.

https://www.spigotmc.org/threads/spigot-bungeecord-1-20-5-1-20-6.645069/#post-4722247

image

### DIFF
--- a/docs/paper/dev/getting-started/plugin-yml.mdx
+++ b/docs/paper/dev/getting-started/plugin-yml.mdx
@@ -90,7 +90,7 @@ This will be shown in the plugin info commands.
 
 ### api-version
 
-The version of the Paper API that your plugin is using. This doesn't include the minor version until 1.20.5. From 1.20.5 and onward a minor version is supported.
+The version of the Paper API that your plugin is using. This doesn't include the minor version until 1.20.5. From 1.20.5 and onward, a minor version is supported.
 Servers with a version lower than the version specified here will refuse to load the plugin.
 The valid versions are 1.13 - <SoftwareVersion versionType={"maj-min-pat"}/>.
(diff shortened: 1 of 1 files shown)


powercasgamer (approved):

zlataovce (approved):

docs/paper/dev/getting-started/plugin-yml.mdx
```diff
@@ -90,7 +90,7 @@ This will be shown in the plugin info commands.
 
 ### api-version
 
-The version of the Paper API that your plugin is using. This doesn't include the minor version.
+The version of the Paper API that your plugin is using. This doesn't include the minor version until 1.20.5. From 1.20.5 and onward, a minor version is supported.
```
rainbowdashlabs, line 93: Looking at it again. Are we really sure to not follow semver here on purpose? The current docs say that the api version has no minor version, although it should be the patch.

I would suggest to actually phrase it like this, which applies more to the correct naming scheme.

```suggestion
The version of the Paper API that your plugin is using. This doesn't include the patch version (e.g. 1.13) until 1.20.5. From 1.20.5 and onward, a patch version (e.g 1.20.5) is supported.
```

zlataovce: I brought this up on the Discord server (#docs-website) yesterday and was told that the community generally ignores the 1 in 1.xx.x, so we should probably stick to calling it `ignored.major.minor` as Minecraft isn't semantically versioned and it's what people expect

granny (approved):

olijeffers0n (approved):

PR #434, opened 2024-06-25
Added note that the docs are incomplete for the Commands API

Hinterhaeltiger:
The Commands API Page in the docs is currently incomplete and/or not detailed especially when it comes to Arguments/Argument types, so I added a note for that while I'm working on expanding the docs

### DIFF
--- a/docs/paper/dev/api/commands.mdx
+++ b/docs/paper/dev/api/commands.mdx
@@ -8,6 +8,14 @@ description: A guide to Paper's Brigadier command API.
 Paper's command system is built on top of Minecraft's Brigadier command system. This system is a
 powerful and flexible way to define commands and arguments.
 
+:::note
+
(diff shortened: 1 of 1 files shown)


olijeffers0n:
Eh, I don't see a particular value in this

Hinterhaeltiger:
> Eh, I don't see a particular value in this

Yeah but as I tried using this doc specifically myself I realized that it is pretty much useless when it comes to implementing more complex commands and arguments so I added a note that links to the example plugin so that people wanting to use the new API have a reference on how to do things. I'm also currently working on expanding the docs so it would just be a note for a few days :)

lynxplay:
I think we are fine without a note :+1: Given the entire thing is experimental anyway, it should be kind of a given things are incomplete. Merging a commit just to revert it later also seems kinda bleh, the brigadier command API is additionally also not something we expect to be 100% easily usable. That is what BasicCommand exists for.

If you want to work on the commands docs, that would be awesome! But yea, I don't see value in adding a note.

PR #444, opened 2024-07-16
Update project-setup to correct JDK version.

Axodouble:
Updated the JDK, as 17 is no longer supported.


Strokkur424:
I am about to do a PR which also updates the picture above

lynxplay:
Just include the picture change in this PR 👍 LGTM otherwise

Axodouble:
> I am about to do a PR which also updates the picture above

Would it not be handier to add it onto this PR?

Strokkur424:
Can I even edit this pr here?

lynxplay:
Just post the picture here, they can include it 👍

Strokkur424:
Very well, here you go:
new-project-paper

lynxplay:
Merci <3

Strokkur424:
Looks Good 👍

docs/paper/dev/getting-started/project-setup.mdx
```diff
@@ -185,7 +185,7 @@ You will be asked to provide some information about your project.
...
 | **Artifact ID**       | The artifact ID of your project. This is used for Maven and Gradle. This is usually the name of your project. This is usually the same as the `Name` field.                                                                                                          |
 | **Version**           | The version of your project. This is used for Maven and Gradle. This is usually `1.0-SNAPSHOT` and does not really matter for now.                                                                                                                                   |
-| **JDK**               | The JDK you want to use. This can be anything from Java 17 and above.                                                                                                                                                                                                |
+| **JDK**               | The JDK you want to use. This can be anything from Java 21 and above.                                                                                                                                                                                                |
```
zlataovce, line 188: this should probably be changed to refer to the `PAPER_JAVA_MIN` property: https://github.com/PaperMC/docs/blob/main/CONTRIBUTING.md#magic-value-handling

Axodouble: Good idea, just committed an attempt for it.

zlataovce: looks good, just trim the excess space from the table field, so the end lines up

Strokkur424:
Looks Good² 👍

MiniDigger:
why does the new screenshot sets the paper-plugin.yml checkbox?
we do not recommend people to use that unless the need it, especially beginners that would use this guide

Strokkur424:
That is a fair point. That was my bad, as I accidentally had it ticked and forgot about it

Issue #413, opened 2024-06-13
Document disable-sprint-interruption-on-attack issue when attacking player


lynxplay:
See https://github.com/PaperMC/Paper/issues/10870, basically a client will locally stop sprinting when attacking a RemotePlayer, leading to a buggy moment of stop + start sprinting.

PR #451, opened 2024-07-20
Document Vanilla Interruption bug

olijeffers0n:
closes https://github.com/PaperMC/docs/issues/413


zlataovce (approved):
is this something that should be noted in the Vanilla-like experience guide? otherwise lgtm

olijeffers0n:
Not sure it needs to be, just because this is just stating that if you change this there will be a vanilla visual bug or whatever.

PR #463, opened 2024-08-13
refactor: Merge vanilla default files with normal config files

Leguan16:
I do not like the current solution to the "how-to vanilla" page. Thats why I am proposing a refactor that merges the two files with adding a `vanilla` field to config options that alter the vanilla behavior. 

```yml
only-merge-items-horizontally:
  vanilla: "true"
  default: "false"
  description: >-
    Prevents merging items that are not on the same y level, preventing potential
    visual artifacts.
```


Leguan16:
should be open for reviews now!

olijeffers0n (changes requested):
I like this implementation a lot more, thanks! One thing, could you please rename this `vanilla` within the ConfigDocBlock to something more like `showVanillaValue` or something slightly more self explanatory. Good to merge then

Leguan16:
Alright will do! I also like it way more than the current version.

Leguan16:
> I like this implementation a lot more, thanks! One thing, could you please rename this `vanilla` within the ConfigDocBlock to something more like `showVanillaValue` or something slightly more self explanatory. Good to merge then

changed the param name.

olijeffers0n (approved):

PR #464, opened 2024-08-13
Document methods to resolve issues with Gradle in Groovy

minacle:
When using `paperweight-userdev` with Gradle and Groovy syntax instead of Gradle Kotlin DSL, some modifications to the code described in the documentation are necessary.

I have added a note to the documentation regarding the issues I encountered and how to resolve them.

### DIFF
--- a/docs/paper/dev/getting-started/userdev.mdx
+++ b/docs/paper/dev/getting-started/userdev.mdx
@@ -148,6 +148,6 @@ This is useful for plugins that have loaders for both Spigot and Paper and want 
 
 :::note
 
-If you are using Gradle with Groovy syntax, you should call the mangled static method like `getMOJANG_PRODUCTION()`.
+If you are using Gradle with the Groovy DSL, you should instead access the fields via static method like `getMOJANG_PRODUCTION()`.
 
 :::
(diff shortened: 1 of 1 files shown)


lynxplay:
Given we don't really provide groovy support anywhere else in the docs (as we *heavily* favour kotlin) I'd be in favour of making this a bit less verbose.

Something like a general note at the bottom saying that "The static members above are accessed with a `get` prefix when using the groovy dsl" 
Or something along those lines. Generally, just wanna make sure it doesn't look like we actually encourage using groovy.

minacle:
How about this?

docs/paper/dev/getting-started/userdev.mdx
```diff
@@ -145,3 +145,9 @@ paperweight.reobfArtifactConfiguration = io.papermc.paperweight.userdev.ReobfArt
 ```
 
 This is useful for plugins that have loaders for both Spigot and Paper and want to keep compatibility with both.
+
+:::note
+
+If you are using Gradle with Groovy syntax, you should call the mangled static method like `getMOJANG_PRODUCTION()`.
```
lynxplay, line 151:
```suggestion
If you are using Gradle with the Groovy DSL, you should instead access the fields via static method like `getMOJANG_PRODUCTION()`.
```

lynxplay:
LGTM to me otherwise :+1:

lynxplay (approved):

lynxplay:
Thank you!

PR #470, opened 2024-08-19
Update audience operation support details in Velocity

Timongcraft:
- Documented support for tablist headers, footers, and resource packs.
- Clarified that unsupported operations include sounds and books. (currently the only unsupported operations)

Support for resource packs was added to the `ConnectedPlayer` class [here](https://github.com/PaperMC/Velocity/commit/cbd07b14349c449202e84a840da1167b42487cca#diff-1cd097c1de92fff8c43bc33eb4f52405eb5396f9e8cc922a4502e4cad12e1f88).
Support for the tablist header and footer was added [here](https://github.com/PaperMC/Velocity/commit/5da085d82f9cd7a91a4b8d44f6dc94dc3383b173#diff-1cd097c1de92fff8c43bc33eb4f52405eb5396f9e8cc922a4502e4cad12e1f88).

Link to the original page: https://docs.papermc.io/velocity/dev/pitfalls#audience-operations-not-supported
Link to the page: https://timongcraft-fix-audience-api.papermc-docs.pages.dev/velocity/dev/pitfalls#audience-operations-are-not-fully-supported

### DIFF
--- a/docs/velocity/dev/getting-started/pitfalls.mdx
+++ b/docs/velocity/dev/getting-started/pitfalls.mdx
@@ -33,5 +33,5 @@ public void onProxyInitialization(ProxyInitializeEvent event) {
 
 ## Audience operations not supported
 
-Velocity only supports sending chat messages, action bar messages, titles, and bossbars through the
-Adventure API. No other operations are supported.
+Velocity only supports sending chat messages, action bar messages, titles, bossbars, tablist header and footer, and resource packs through the
+Adventure API. All other operations, such as sound and book operations, are not supported.
--- a/docs/velocity/dev/getting-started/pitfalls.mdx
+++ b/docs/velocity/dev/getting-started/pitfalls.mdx
@@ -33,6 +33,9 @@ public void onProxyInitialization(ProxyInitializeEvent event) {
 
 ## Audience operations are not fully supported
 
+Velocity currently does not support all Audience operations of the Adventure API, so these operations should be handled on the backend.
+Furthermore, playing sound was previously considered infeasible, as for versions below 1.19.3, a registry of hardcoded sound IDs is required.
+
 | Operation                  | Supported |
 | -------------------------- | ----------|
 | Chat messages              | Yes       |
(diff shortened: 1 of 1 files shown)


docs/velocity/dev/getting-started/pitfalls.mdx
```diff
@@ -33,5 +33,5 @@ public void onProxyInitialization(ProxyInitializeEvent event) {
 
 ## Audience operations not supported
 
-Velocity only supports sending chat messages, action bar messages, titles, and bossbars through the
-Adventure API. No other operations are supported.
+Velocity only supports sending chat messages, action bar messages, titles, bossbars, tablist header and footer, and resource packs through the
```
kashike, line 36: Is it worth having a table here instead?

Timongcraft: We could, but if Kiory adds something, it won't be listed and would need to be updated.

Timongcraft: *Literal* or with a component?
I see both in the docs, so I'm not sure what I should use here.

zlataovce (approved):

Leguan16:
The table alone after a heading looks weird. I think some sort of introduction to why some operations are not supported would be nice to have.

Due to <reason> and <reason> Velocity currently only supports the following audience operations:
<table>

Timongcraft:
> The table alone after a heading looks weird. I think some sort of introduction to why some operations are not supported would be nice to have.
> 
> Due to and Velocity currently only supports the following audience operations:

You mean a general introduction before the table, right, and not a description for each operation?

Leguan16:
> > The table alone after a heading looks weird. I think some sort of introduction to why some operations are not supported would be nice to have.
> 
> > 
> 
> > Due to and Velocity currently only supports the following audience operations:
> 
> 
> 
> You mean a general introduction before the table, right, and not a description for each operation?

Yes just a message before the table explaining why there are certain methods not implemented.

zlataovce (approved):

PR #473, opened 2024-08-25
The misc.max-leash-distance is now a DoubleOr.Default

lynxplay:
See: papermc/paper#11301


powercasgamer (commented):
Should this mention what the default is?

lynxplay:
I don't think we have been doing that yet, because we don't have a good way to keep that value up to date.

zlataovce (approved):

PR #474, opened 2024-08-26
Update forwarding.md

BiraruYT:


olijeffers0n (approved):

olijeffers0n:
Generally we don’t bother with such small changes. Just something to note going forwards.

BiraruYT:
My bad, thanks for telling > Generally we don’t bother with such small changes. Just something to note going forwards.

PR #475, opened 2024-08-26
Add docs for new alternate-current config

Machine-Maker:
Docs for https://github.com/PaperMC/Paper/pull/11333


olijeffers0n:
Should we specify what each of these mean?

PR #479, opened 2024-09-03
set event result, and add warnings to velocity plugin messaging docs

456dev:
as well as the warnings and updated example code, it also includes 
- more specific diagrams for each case
- fixed identifier checking, rather than by identity
- instanceof style cast.

### DIFF
--- a/docs/velocity/dev/api/plugin-messaging.mdx
+++ b/docs/velocity/dev/api/plugin-messaging.mdx
@@ -17,6 +17,20 @@ flowchart LR
     backend -->|"3 (Incoming)"| Velocity -->|"4 (Outgoing)"| player
 ```
 
+:::warning
+
+When listening to PluginMessageEvent, ensure sure the result is
+<Javadoc name={"com/velocitypowered/api/event/connection/PluginMessageEvent$ForwardResult#handled()"} project={"velocity"}>`ForwardResult.handled()`</Javadoc>
+if you do not intend the client to participate.
+
+If the result is forwarded, Players can spoof the proxy to your backends.
+
+Additonally, ensure the result is set correct after actually handling correct messages, to prevent them from being leaked to the other party.
+
+This can be achieved with unconditionally setting the result between checking the identifier and checking the source, as shown in the examples.
+
+:::
+
 Additionally, BungeeCord channel compatibility is included, which may remove the need for a companion Velocity plugin in certain cases.
 
 ## Case 1: Receiving a plugin message from a player
@@ -25,6 +39,19 @@ This is for when you need to handle or inspect the contents of a plugin message 
 It will require registering with the ChannelRegistrar for the event to be fired.
 
 An example use case could be logging messages from a mod that reports the enabled features.
+
+```mermaid
+flowchart LR
+  subgraph handle from player
+    direction LR
+    a[player] --> b[Velocity] ~~~ c[backend]
+  end
+  subgraph forward from player
+    direction LR
+    d[player] --> e[Velocity] --> f[backend]
+  end
+```
+
 
 ```java
 public static final MinecraftChannelIdentifier IDENTIFIER = MinecraftChannelIdentifier.from("custom:main");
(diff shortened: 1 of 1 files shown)


docs/velocity/dev/api/plugin-messaging.mdx
```diff
@@ -17,6 +17,20 @@ flowchart LR
     backend -->|"3 (Incoming)"| Velocity -->|"4 (Outgoing)"| player
 ```
 
+:::warning
+
+When listening to PluginMessageEvent, ensure sure the result is
```
zlataovce, line 22:
```suggestion
When listening to `PluginMessageEvent`, ensure the result is
```

docs/velocity/dev/api/plugin-messaging.mdx
```diff
@@ -17,6 +17,20 @@ flowchart LR
...
+<Javadoc name={"com/velocitypowered/api/event/connection/PluginMessageEvent$ForwardResult#handled()"} project={"velocity"}>`ForwardResult.handled()`</Javadoc>
+if you do not intend the client to participate.
+
+If the result is forwarded, Players can spoof the proxy to your backends.
+
+Additionally, ensure the result is set correct after actually handling correct messages, to prevent them from being leaked to the other party.
```
zlataovce, line 28:
```suggestion
Additionally, ensure the result is set correctly after actually handling correct messages, to prevent them from being leaked to the other party.
```

docs/velocity/dev/api/plugin-messaging.mdx
```diff
@@ -36,12 +63,23 @@ public void onProxyInitialization(ProxyInitializeEvent event) {
 
 @Subscribe
 public void onPluginMessageFromPlayer(PluginMessageEvent event) {
-    if (!(event.getSource() instanceof Player)) {
+    // Check if Identifier matches first, no matter the source.
```
zlataovce, line 66:
```suggestion
    // Check if the identifier matches first, no matter the source.
```

docs/velocity/dev/api/plugin-messaging.mdx
```diff
@@ -36,12 +63,23 @@ public void onProxyInitialization(ProxyInitializeEvent event) {
...
-    if (event.getIdentifier() != IDENTIFIER) {
+
+    // mark PluginMessage as handled, indicating that the contents
+    // should not be forwarding to their original destination.
+    event.setResult(PluginMessageEvent.ForwardResult.handled());
+
+    // Alternatively:
+
+    // mark PluginMessage as forwarded, indicating that the contents
+    // should be passed through, as if velocity is not present.
```
zlataovce, line 78:
```suggestion
    // should be passed through, as if Velocity is not present.
```

docs/velocity/dev/api/plugin-messaging.mdx
```diff
@@ -114,12 +173,28 @@ public void onProxyInitialization(ProxyInitializeEvent event) {
 
 @Subscribe
 public void onPluginMessageFromBackend(PluginMessageEvent event) {
-    if (!(event.getSource() instanceof ServerConnection)) {
+    // Check if Identifier matches first, no matter the source.
```
zlataovce, line 176:
```suggestion
    // Check if the identifier matches first, no matter the source.
```

docs/velocity/dev/api/plugin-messaging.mdx
```diff
@@ -17,6 +17,20 @@ flowchart LR
     backend -->|"3 (Incoming)"| Velocity -->|"4 (Outgoing)"| player
 ```
 
+:::warning
+
+When listening to PluginMessageEvent, ensure sure the result is
+<Javadoc name={"com/velocitypowered/api/event/connection/PluginMessageEvent$ForwardResult#handled()"} project={"velocity"}>`ForwardResult.handled()`</Javadoc>
+if you do not intend the client to participate.
+
+If the result is forwarded, Players can spoof the proxy to your backends.
```
zlataovce, line 26:
```suggestion
If the result is forwarded, players can impersonate the proxy to your backend servers.
```

zlataovce (changes requested):
need someone familiar with Velocity to check whether the logic is correct, not familiar enough with it myself

docs/velocity/dev/api/plugin-messaging.mdx
```diff
@@ -17,6 +17,20 @@ flowchart LR
     backend -->|"3 (Incoming)"| Velocity -->|"4 (Outgoing)"| player
 ```
 
+:::warning
```
zlataovce, line 20: this admonition seems out of place

456dev: would it be better to be below the mention of the bungeecord plugin channel compat?

as this is probably the most important information on the page, security wise, i would prefer it to remain above the fold, as well as remain emphasized.

zlataovce: after some deliberation, I initially thought it'd be better in Case 1, as the first section is more abstract and not really about handling the messages, but then it'd need to be reiterated in Case 3, so it's probably fine as it is

you can make the admonition `danger` for additional emphasis

zlataovce (approved):

PR #484, opened 2024-09-30
chore: update docusaurus to 3.5.2 and migrate tsconfig for docusaurus 

Doc94:
This PR just update docusaurus to 3.5.2 and also migrate the devPackage for the tsconfig used by docusaurus.
the tsconfig currently used is for the v2 of docusaurus and not the v3.

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,16 +16,16 @@
     "format:check": "prettier . --check"
   },
   "dependencies": {
-    "@docusaurus/core": "3.5.1",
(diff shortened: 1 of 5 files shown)


Doc94:
PS: First time with pnpm and the "patch" i test in local and run.. i hope not forget any for update the patch (i just load the patch and commit again with the same changes)

zlataovce (approved):

PR #490, opened 2024-10-15
Document session and services host properties

Phoenix616:
This adds documentation for the `minecraft.api.session.host` and `minecraft.api.services.host` system properties.


Leguan16:
Are these the only system properties provided by Minecraft?

zlataovce (approved):

Phoenix616:
> Are these the only system properties provided by Minecraft?

Well I didn't look everywhere (they aren't exactly in the same place in the source code...) and in this area there is only an additional one to switch to the staging URLs of the official Mojang API servers via a single system property which don't seem to be useful in the Paper context hence why I decided to not include that. (And if you need those URLs you can set them manually via the documented values anyways)

PR #492, opened 2024-10-25
Make userdev version be fetched dynamically

Strokkur424:
Instead of hardcoding a version value or focing the user to fetch it itself (previous change to this site),  it should just be fetched live.

I have no idea what I am doing, so suggest changes freely


Strokkur424:
No idea what the format issue is. If somebody can tell me, please do so

zlataovce:
I'm not particularly fond of creating another component just for userdev, I'd much rather have it be a part of `VersionFormattedCode` (+ the placeholder should be documented in CONTRIBUTING.md)

Strokkur424:
@zlataovce actually thinking about it, maybe I should put that util method into util/versionUtils.ts. Give me a second

zlataovce:
> @zlataovce actually thinking about it, maybe I should put that util method into util/versionUtils.ts. Give me a second

yes, it should be cached in an expiring value

zlataovce (approved):

src/util/versionUtils.ts
```diff
@@ -21,7 +21,6 @@ class ExpiringValue<T> {
     return this._value;
   }
 }
-
```
powercasgamer, line 24: undo

powercasgamer (changes requested):
otherwise looks *fine*

src/util/versionUtils.ts
Strokkur424: there

powercasgamer (approved):

olijeffers0n (approved):

PR #499, opened 2024-11-25
Add common sense.

ProdPreva1l:
Adds a page at /misc/common-sense which is a nicely worded message telling someone they need some help mentally.


zlataovce:
I don't see the informational value of having such a page, to me it mainly comes off as rude if I were the recipient of such a link

ProdPreva1l:
> I don't see the informational value of having such a page, to me it mainly comes off as rude if I were the recipient of such a link

thats kinda the point

PR #500, opened 2024-11-25
Bump minimum velocity version to 21

AoElite:
Bump minimum velocity version to 21. It appears intentional that Velocity is ran with at least Java 21 since parts of the code base rely on specific Java 21 features. Such as commit https://github.com/PaperMC/Velocity/commit/ecf936f35665f9fd0f65ee114062cfbda2b89bf6.

### DIFF
--- a/config-specs/properties.json
+++ b/config-specs/properties.json
@@ -1,5 +1,5 @@
 {
   "DOCS_JAVA": "21",
   "PAPER_JAVA_MIN": "21",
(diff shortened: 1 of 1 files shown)


zlataovce:
that commit looks like it was made to be compatible with versions <21 and the defined toolchain is still Java 17, so I'd appreciate it if someone from the Velocity team could weigh in here (cc @4drian3d)

maybe it'd be worth to specify a different recommended version beside the minimum version

PR #505, opened 2024-12-12
feat(misc/assets): Add Hangar branding guidelines

MC-XiaoHei:
[in discord](https://discord.com/channels/289587909051416579/1098348646757568541/1316729762995241002)


zlataovce:
as there are no existing guidelines (to my knowledge), this needs a go-ahead from the core/Hangar team

otherwise seems to be fine, considering the logo mark looks to be a derivative of the Paper one

Leguan16:
Is there a reason why it's between Paper and Velocity and not after Velocity? Just curious. 😄

MC-XiaoHei:
> Is there a reason why it's between Paper and Velocity and not after Velocity? Just curious.

It doesn't make any sense, it's just that I think the icon of papermc is more similar to hanger.

PR #508, opened 2024-12-16
feat: swap to a version tag rather than strict versioning

olijeffers0n:


powercasgamer (changes requested):
need to handle redirects

zlataovce:
superseded by #560

PR #514, opened 2024-12-29
Feat/data components

Owen1212055:
Please feel free to make modifications directly to the PR.

This is meant as an initial base.


Strokkur424 (commented):
Definitely helps a ton with understanding data components.
One thing I would maybe suggest to add is a small line which explains whether you can use ItemMeta and DataComponents concurrently, since I am not quite sure about that, even after reading your docs.

EDIT: Also sorry if some of these reviews are broken. I haven't really used this specific feature of GitHub yet

Owen1212055:
I'd rather not be a blocker here, please feel free to checkout this branch and make modifications. :) 
I sadly cannot dedicate much time to fix this, but i'd like to get this through in some degree.

zlataovce (approved):

PR #520, opened 2025-01-09
Updated chmod permission

Natfan:
Within Linux's chmod system, `+x` grants all users access to execute the file, whereas `u+x` only grants the user who owns the file access to execute it. Provided the user is the one who created the `start.sh` file, there's no reason why this permission change wouldn't cause issues, but it's good to get into the habit of using "least privilege" when assigning permissions, and I know from personal experience that hosting Minecraft servers can be the start of a career in sysadmining.

### DIFF
--- a/docs/velocity/admin/getting-started/README.mdx
+++ b/docs/velocity/admin/getting-started/README.mdx
@@ -59,7 +59,7 @@ java -Xms1G -Xmx1G -XX:+UseG1GC -XX:G1HeapRegionSize=4M -XX:+UnlockExperimentalV
 
 Once saved, open a terminal (or log into the machine) if you haven't already, navigate to the
 directory where you have placed the Velocity JAR file and the `start.sh` file. Then run
-`chmod +x start.sh` and then `./start.sh`. If it worked, you should now receive a console similar to
+`chmod u+x start.sh` and then `./start.sh`. If it worked, you should now receive a console similar to
 the output in the next section.
 
 ## After launch


MiniDigger:
I like the idea of encouraging better habits, maybe we even should actually explain what the command does, i.e. "then run chmod to make the file executable"

docs/velocity/admin/getting-started/README.mdx
```diff
@@ -59,7 +59,7 @@ java -Xms1G -Xmx1G -XX:+UseG1GC -XX:G1HeapRegionSize=4M -XX:+UnlockExperimentalV
 
 Once saved, open a terminal (or log into the machine) if you haven't already, navigate to the
```
olijeffers0n, line 60: I like the idea of this PR, maybe a paragraph like this works better though?

---

Once you've saved the files, open a terminal (or log into the machine) if you haven't already. Navigate to the directory where you have placed the Velocity JAR file and the `start.sh` file. Then, you will need to prepare the start.sh script to be executable and run it.

1) Run `chmod u+x start.sh`. This command modifies the file's permissions, granting the file's owner (you) the ability to execute (x) the script. Without this step, the system may not recognize the script as something that can be run.

2) Now that the file is executable, run `./start.sh`  to run the script.

olijeffers0n (changes requested):

docs/velocity/admin/getting-started/README.mdx
Strokkur424: Since the PR is stale, I decided to apply your suggestion.

Strokkur424 (approved):

Issue #512, opened 2024-12-23
Start Script Generator


NonSwag:
_Didn't know what to name this issue, feel free to change it_

The start script generator doesn't use the recommended values for >=12GB of RAM
  
    -XX:G1NewSizePercent=40
    -XX:G1MaxNewSizePercent=50
    -XX:G1HeapRegionSize=16M
    -XX:G1ReservePercent=15
    -XX:InitiatingHeapOccupancyPercent=20

source  [Aikar's blog: If you are using an Xmx value greater than 12G](https://aikar.co/2018/07/02/tuning-the-jvm-g1gc-garbage-collector-flags-for-minecraft/)

zlataovce:
I believe the 12G+ tweaks are not recommended for modern Minecraft versions as they cause more GC issues than they solve, so that's why they're not used nor mentioned on the Aikar's flags page

PR #547, opened 2025-03-02
Fixed Aikar's flags in "Start Script Generator"

ZepsiZola:
Some of the flags in Aikar's Flags change slightly depending on how much RAM you're allocating to the server. [(Source)](https://aikar.co/category/minecraft/#:~:text=If%20you%20are%20using%20an%20Xmx%20value%20greater%20than%2012G)

The "[Start Script Generator](https://docs.papermc.io/misc/tools/start-script-gen)" on PaperMC's docs does not currently reflect this.

I have made changes to `StartScriptGenerator.tsx` to properly output Aikar's flags depending on the RAM.

The changes were:
- Added `altFlags` to the `FLAGS` `Option[]` in `StartScriptGenerator.tsx` to define flags that change depending on `memory`.
- Creation of the `getAltFlags(memory, selectedFlag)` function in`StartScriptGenerator.tsx`, which returns a string of the correct flags according to the provided `memory` and `selectedFlag`. Returns "" if `altFlags` is not present.
- Called `${getAltFlags(memory, selectedFlag)}` in the `const command = "..."` part of `StartScriptGenerator.tsx`

I originally also changed `Option` interface defined in `Select.tsx`, but then I realised my changes in `StartScriptGenerator.tsx` worked regardless of if I added the line `altValue?: Map<number, string>;` to the defined Option interface, so I kept it as it was. Feel free to tell me if I implemented this part wrong.

I tried to make this as non-spaghetti code as possible. The memory boundaries for `altFlags` are easily adjustable and you can easily add new `altFlags` entries for other sets of flags should it ever be necessary.

## Modified "Start Script Generator" working:
unknown
unknown

### DIFF
--- a/src/components/StartScriptGenerator.tsx
+++ b/src/components/StartScriptGenerator.tsx
@@ -32,7 +32,6 @@ const FLAGS: Option[] = [
       "-XX:G1MixedGCCountTarget=4",
       "-XX:G1MixedGCLiveThresholdPercent=90",
       "-XX:G1RSetUpdatingPauseTimePercent=5",
-      "-XX:InitiatingHeapOccupancyPercent=15",
       "-XX:MaxGCPauseMillis=200",
       "-XX:MaxTenuringThreshold=1",
       "-XX:SurvivorRatio=32",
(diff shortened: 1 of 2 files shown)


Leguan16:
Afaik the section about 12g+ is outdated and its ok to use the normal flags for higher ram values.

zlataovce:
thank you for the PR, however the 12G tweaks are intentionally not included, as explained [here](https://github.com/PaperMC/docs/issues/512#issuecomment-2560203098)

(also, GC-wise, modern JVMs work very well with default settings, so Aikar's tweaks aren't as important as they used to be a few years ago)

ZepsiZola:
Damn.

PR #549, opened 2025-03-11
Update intro.mdx

Ttempor:
This partial code is hard to understand，needs to be completed。


electronicboy:
> Components are designed for use with static method imports

MiniDigger:
maybe we could add an example static import, but your IDE should recommend it and people should know what static imports are

PR #557, opened 2025-04-07
feat: move userdev advise for gradle version to note block

Doc94:
Close https://github.com/PaperMC/docs/issues/556 moving the mention of the gradle version to a block note for make more clear about the need of keep gradle updated.

PR #558, opened 2025-04-07
Make Gradle update-version block more visible

### LINKED ISSUE #556: Make userdev requiring latest gradle versions a warning block
We have a large increase in users running into the issue that userdev requires gradle 8.12 or above.
The notice for that is rather simple to read over on the docs page as it is part of a larger text block.
It might be worth highlighting this in an alert, e.g. warning or note.

Strokkur424:
Fixes #556


zlataovce:
Superseded by #557.

PR #564, opened 2025-04-13
Change the homepage cards to be clickable

olijeffers0n:
New:
image
Before:
image


Doc94:
the before not was like this?
image

or i missing a update?

olijeffers0n:
They were vertically scattered… not sure why we have different views though.

zlataovce:
> They were vertically scattered… not sure why we have different views though.

I changed it in the latest commit

zlataovce (approved):

PR #568, opened 2025-04-13
feat: Item Converter [Starlight]

Strokkur424:


Strokkur424:
The actual converter logic is currently commented out because it throws errors. Somebody has to look at that, as I do not know what to change.

But you can already review the UI. That's basically the biggest part of this PR anyways

PR #575, opened 2025-04-20
feat: Add last modified by

Strokkur424:
This re-adds the previous behavior where it also displays the person who last edited a certain page. For obvious reasons, this mostly displays only @zlataovce. But I have also verified that on the pages `/paper/basic-troubleshooting/` and `/paper/reference/bukkit-help-configuration/`, which have been modified since the starlight rewrite, the correct PR author is noted down.

There is a known bug where it shows the correct name for local/branched commits, but shows an incorrect link. This does not impact production builds though. The account link is, if the email cannot be traced back to an account directly, retrieved from the latest commit on a file from the main branch of the repository, which may be outdated.

### DIFF
--- a/src/components/overrides/LastUpdated.astro
+++ b/src/components/overrides/LastUpdated.astro
@@ -1,0 +1,24 @@
+---
+import { type GitHubAccount, getGitHubAccountFromFile } from "/src/utils/git-utils.ts"
+
+const { lang, lastUpdated } = Astro.locals.starlightRoute;
+const filePath = Astro.locals.starlightRoute.entry.filePath
+
+const acc: GitHubAccount = await getGitHubAccountFromFile(filePath);
+---
+
+{
+  lastUpdated && (
+    <p>
+      {"Last updated: "}
+      <time datetime={lastUpdated.toISOString()}>
+        {lastUpdated.toLocaleDateString(lang, {
+          dateStyle: "medium",
+          timeZone: "UTC",
+        })}
+      </time>
+      {" by "}
+      <a set:html={acc.displayName} href={acc.accountLink} />
+    </p>
+  )
+}
--- a/src/content/docs/paper/dev/api/command-api/basics/introduction.md
+++ b/src/content/docs/paper/dev/api/command-api/basics/introduction.md
@@ -48,4 +48,4 @@ The following pages will be added to the documentation in the future:
 :::
 
 ## Additional support
-For support regarding the command API, you can always ask in our [Discord Server](https://discord.gg/PaperMC) in the `#paper-dev` channel!
+For support regarding the command API, you can always ask in our Discord server under the [`#paper-dev`](https://discord.com/channels/289587909051416579/555462289851940864) channel!
--- a/src/utils/git-utils.ts
+++ b/src/utils/git-utils.ts
@@ -1,0 +1,79 @@
+import { execSync } from "child_process";
+
+export interface GitHubAccount {
+  displayName: string;
+  email: string;
+  accountLink?: string;
+}
+
+const headers = {
+  headers: {
+    Accept: "application/vnd.github+json",
+    "User-Agent": "papermc-docs/author",
+  },
+};
+
+const repo: string = "PaperMC/docs"
+const cache: Map<string, GitHubAccount> = new Map();
+
+// Git
+export async function getGitHubAccountFromFile(
+  filePath: string
+): Promise<GitHubAccount | null> {
+  const displayName = execSync(
+    `git log -1 --pretty="format:%an" -- "${filePath}"`
+  ).toString();
+  const email = execSync(
+    `git log -1 --pretty="format:%ae" -- "${filePath}"`
+  ).toString();
+
+  const cached = cache.get(email);
+  if (cached != null) {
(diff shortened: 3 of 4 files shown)


src/content/docs/paper/dev/api/command-api/basics/introduction.md
```diff
@@ -48,4 +48,4 @@ The following pages will be added to the documentation in the future:
 :::
 
 ## Additional support
-For support regarding the command API, you can always ask in our [Discord Server](https://discord.gg/PaperMC) in the `#paper-dev` channel!
+For support regarding the command API, you can always ask in our Discord server under the [`#paper-dev`](https://discord.com/channels/289587909051416579/555462289851940864) channel!
```
Doc94, line 51: Try add the two url because if you are not in the server the channel url just throw a not found issue.

src/components/overrides/LastUpdated.astro
```diff
@@ -0,0 +1,24 @@
...
+---
+
+{
+  lastUpdated && (
+    <p>
+      {"Last updated: "}
+      <time datetime={lastUpdated.toISOString()}>
+        {lastUpdated.toLocaleDateString(lang, {
+          dateStyle: "medium",
+          timeZone: "UTC",
```
Doc94, line 17: You can not make this depend in the browser timezone?

zlataovce: that's just how starlight does it

src/content/docs/paper/dev/api/command-api/basics/introduction.md
Strokkur424: Did you test that? I am under the assumption that, due to PaperMC being a findable guild, it would work fine

src/components/overrides/LastUpdated.astro
Strokkur424: It should matter too much, but it now does that

src/content/docs/paper/dev/api/command-api/basics/introduction.md
Strokkur424: Yeah I tested it just now. The link works fine, even if you haven't joined the Discord server before.
image

granny: Does not work for guilds, even if they are public. You have to join the guild first for it to work.

It will show this to people who aren't in the PaperMC guild when they click the link:

image

(Discord truncates the channel id from the url after loading)

zlataovce: this diff should just be reverted if it's for testing

Doc94:
Maybe is more easy just show the author with not an url (like old docs) and just include the URL for commit to any how wants go to the commit in github and then go to the profile... maybe with that its not necesary use the github API

granny (changes requested):

src/utils/git-utils.ts
```diff
@@ -0,0 +1,54 @@
...
+// Git
+export async function getGitHubAccountFromFile(filePath: string): Promise<GitHubAccount | null> {
+  const email = execSync(`git log -1 --pretty="format:%ae" -- "${filePath}"`).toString();
+  const cached = emailCache.get(email);
+  if (cached != null) {
+    return cached;
+  }
+
+  const displayName = execSync(`git log -1 --pretty="format:%an" -- "${filePath}"`).toString();
+  const hash = execSync(`git log -1 --pretty="format:%H" -- "${filePath}"`).toString();
```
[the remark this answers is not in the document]
zlataovce: GitHub API can get ratelimited easily (when unauthenticated), so this is a good fallback

src/utils/git-utils.ts
```diff
@@ -0,0 +1,54 @@
...
+          "User-Agent": "papermc-docs/author",
+        },
+      };
+
+const repo: string = "PaperMC/docs";
+const emailCache: Map<string, GitHubAccount> = new Map();
+
+// Git
+export async function getGitHubAccountFromFile(filePath: string): Promise<GitHubAccount | null> {
+  const email = execSync(`git log -1 --pretty="format:%ae" -- "${filePath}"`).toString();
```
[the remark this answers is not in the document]
zlataovce: where do you get the username at that point though? the committer's name doesn't always correspond to the GitHub username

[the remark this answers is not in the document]
granny: right.. i wish github exposed the github display name in the commit endpoint :/ you're able to grab it through the username endpoint, but that would be an extra request if used here, which isn't really worth it

src/components/overrides/LastUpdated.astro
```diff
@@ -0,0 +1,30 @@
...
+        })}
+      </time>
+      {info && (
+        <>
+          {" by "}
+          <a href={info.committer.href}>{info.committer.name}</a>
+          {" in "}
+          <a href={`https://github.com/${REPO}/commit/${info.hash}`}>{info.hash.substring(0, 7)}</a>
```
Doc94, line 25: Maybe better get the two hashs and not make a substring? (With two hashes i mean the large and short)

zlataovce (approved):

PR #578, opened 2025-04-21
Fix userdev page

Strokkur424:
This PR introduces two QoL changes on the userdev page:
- Adds dynamic latest version fetching (the same as it had previously).
- Fixes syntax highlighting of the `version` label.

The syntax highlight is fixed by having a smaller, custom syntax highlight definition made for exactly that one code block.

The dynamic version fetching is done using GitHub's tag API on the paperweight repo


zlataovce:
dynamic version for userdev looks fine, but adding a grammar specifically for a singular code block is overkill

Strokkur424:
I personally think that it adds a lot to the feel of the page. If you want to, I could modify the actual Kotlin syntax file so that it is included automatically, though that would result in more complexity of the diff and thus result in a higher maintenance cost

zlataovce:
changing the Kotlin grammar to properly highlight infix functions would be the better solution, but that seems like a PR for shiki/tm-grammars and not this site, so I am against including it

olijeffers0n (changes requested):
I agree with @zlataovce - this just adds bloat to the repository for zero benefit really. Code for automatic version grabbing is good to ship, however.

src/content/docs/paper/dev/getting-started/userdev.md
```diff
@@ -36,9 +36,9 @@ See [here](#1205-and-beyond) for more details.
 ## Adding the plugin
 Add the plugin to your `build.gradle.kts` file.
 
-```kotlin title="build.gradle.kts"
+```kts title="build.gradle.kts" replace
```
olijeffers0n, line 39: I'm assuming kts is a synonym.

olijeffers0n (approved):

zlataovce (approved):

Doc94:
i know this was merged but maybe was better check in the metadata of [gradle plugin](https://plugins.gradle.org/) rather than tags created in the repo [ref](https://github.com/badges/shields/blob/ea559e7d8b4cbfc432aac37e1d5aad1beacc7802/services/gradle-plugin-portal/gradle-plugin-portal.service.js#L36).

zlataovce:
parsing Maven meta (Gradle plugin portal is a Maven repo iirc) isn't exactly fun (XML) and paperweight has proper Git tags, so it should work just fine

PR #589, opened 2025-05-01
feat: add xp-merge-percentage config

mja00:
Documents config value created in https://github.com/PaperMC/Paper/pull/12503


lynxplay (approved):

zlataovce:
it should probably mention the default value (40) in the description, so people changing the configuration option have a reference value without having to dig through Paper patches

lynxplay:
Yea, I guess we can also mention the behaviour around value equality etc.

lynxplay (approved):

PR #590, opened 2025-05-08
chore: housekeeping

### LINKED ISSUE #587: paper-world-defaults.yml's door-breaking-difficulty field doesn't match the one created by paper server
https://github.com/PaperMC/docs/blob/79db3f777af4f881e65f657a7219ee9fbfcf1a38/src/config/paper/paper-world-defaults.yml#L201-L205

Which creates the following:
```yml
door-breaking-difficulty:
 - HARD
 - NORMAL
 - EASY
 - PEACEFUL
```

But the paper server generated one looks as follows:
```yml
door-breaking-difficulty:
  husk:
  - HARD
  vindicator:
  - NORMAL
  - HARD
  zombie:
  - HARD
  zombie_villager:
  - HARD
  zombified_piglin:
  - HARD
```

@Lulu13022002 mentioned in discord that it would be best to just mimic the `despawn-time` section

zlataovce:
Closes #587.


Doc94:
feel strange use two texts for the experimental flag... can be confusing maybe just using a emote related like?
image

also can be great if the badge allow title/alt-text but i think that its not supported not?

zlataovce:
> feel strange use two texts for the experimental flag... can be confusing maybe just using a emote related like? image
> 
> also can be great if the badge allow title/alt-text but i think that its not supported not?

an emoji doesn't look good imo; I will just make them all "Experimental", so it's consistent

olijeffers0n (approved):
image

This sidebar is quite a lot with the experimental tags, but not a blocker for me.

zlataovce:
> image
> 
> This sidebar is quite a lot with the experimental tags, but not a blocker for me.

I thought it'd make sense to have all these tagged if we already do that with Paper plugins

PR #573, opened 2025-04-20
feat: Add Adventure docs

Strokkur424:
Redo of #569.

PR #596, opened 2025-05-20
chore: use node 22 and github actions improvements

Doc94:
Related to https://github.com/PaperMC/docs/pull/573 this PR try to make more clear what node engine version need to use also improvement the github action by using compose for a few "duplicate" steps.

### DIFF
--- a/.github/workflows/spellcheck.yaml
+++ b/.github/workflows/spellcheck.yaml
@@ -14,4 +14,4 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v4
       - name: Spell check
-        uses: crate-ci/typos@v1.24.6
+        uses: crate-ci/typos@v1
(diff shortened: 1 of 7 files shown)


.github/workflows/spellcheck.yaml
```diff
@@ -14,4 +14,4 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v4
       - name: Spell check
-        uses: crate-ci/typos@v1.24.6
+        uses: crate-ci/typos@v1
```
zlataovce, line 17: we don't want to target a wide range, as it may spontaneously start failing CI runs due to a previously unchecked mistake

Doc94: Okay i change for the specify latest based in how Action dont throw any with that version `[1.32.0](https://github.com/crate-ci/typos/releases/tag/v1.32.0)`

zlataovce (approved):

PR #600, opened 2025-05-28
Add a note saying `sync-chunk-writes` is noop

Toffikk:


src/config/paper/server-properties.yml
```diff
@@ -225,7 +225,9 @@ spawn-protection:
     at least 1 operator to be enabled.
 sync-chunk-writes:
   default: "true"
-  description: "Enables synchronous writing of chunk data."
+  description: |
+    Enables synchronous writing of chunk data. Has no effect on Paper by default,
```
olijeffers0n, line 229: Maybe it would be nice to specify if one takes precedence over the other, or does this one just not do anything? Could be clarified I think.

Toffikk: Modified

PR #604, opened 2025-06-07
chore: misc styling touch-ups

zlataovce:
- makes some light mode grays lighter for better legibility of black text on gray backgrounds
- makes the config node highlight full width to not mess up indentation (+ accent color instead of gray so it's easier to see on light backgrounds)
- moves the scrolling area of the sidebar to not include the dropdown (no longer a sticky element)

### DIFF
--- a/src/components/ConfigNode.astro
+++ b/src/components/ConfigNode.astro
@@ -59,7 +59,7 @@ const formatDefault = (value?: string): Default => {
           {"message" in rawValue ? (
             <div
               class:list={["block", "message", rawValue.inline && "message-inline"]}
-              style={rawValue.color ? `border-color: ${rawValue.color}` : undefined}
+              style={rawValue.color ? `border-left-color: ${rawValue.color}` : undefined}
               set:html={render(rawValue.message)}
             />
           ) : "description" in rawValue ? (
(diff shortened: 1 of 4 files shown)


Strokkur424 (changes requested):
A few things I have noticed, which are a bit weird (could also be in the production docs, but could still be useful to fix here):

- When switching dark/light mode, the project boxes in the hero have a transition effect on them. This might be intended, but it is a bit inconsistent with the rest of the application:
  
  https://github.com/user-attachments/assets/7317dd7a-22da-42ab-ad12-f37904bc6b0c
  
- In light mode, hovering over the sites in the navigation bar at the left does not yield any visible highlighting, which is inconsistent to dark mode behavior:

  https://github.com/user-attachments/assets/b0018437-9739-433d-a56f-5915d72b14c0

- And finally, you cannot scroll down on the nav bar. This might be possible via scroll wheel (I wouldn't know, as my scroll wheel broke when editing moonrise patches), but there neither no scroll bar nor can you scroll down when pressing the middle-mouse button, which is bad for accessibility.

  https://github.com/user-attachments/assets/282e9934-967d-43b1-a37f-1e8a0572e362

zlataovce:
> When switching dark/light mode, the project boxes in the hero have a transition effect on them. This might be intended, but it is a bit inconsistent with the rest of the application

I don't think this is possible to fix without involving JS logic, since moving the transition to the :hover selector would mean losing the ease out animation on hover out

> In light mode, hovering over the sites in the navigation bar at the left does not yield any visible highlighting, which is inconsistent to dark mode behavior

not sure what browser you're trying it on, but I see a color change on hover on both Firefox and Chrome, albeit faint - bumped up the black tone to be more pronounced

Strokkur424:
The scrollbar being gone is now fixed, thanks for that. I can now also notice a faint color change, but it could perhaps be more visible if you also bumped up the font weight a bit. Because it is still very very faint

Strokkur424 (approved):
I also think it looks good

PR #618, opened 2025-07-01
feat: stable responsive images

darbyjack:
This PR will enable responsive image generation during build-time. It is a global setting that can be overridden on a per-image basis, but, by default supports all images inside of MD and MDX files.

### DIFF
--- a/astro.config.ts
+++ b/astro.config.ts
@@ -442,7 +442,7 @@ export default defineConfig({
   },
   image: {
     responsiveStyles: true,
-    layout: 'constrained',
+    layout: "constrained",
   },
   markdown: {
     remarkPlugins: [


Strokkur424:
What does that change? Could you point me to a page where this affects something?

electronicboy:
The real advantage is that you don't notice anything, but, if you see the anti-xray page, the image there on mobile devices, with a smaller screen resolution, is much smaller than on your desktop browser width as it doesn't need to pull in the full resolution image immediatly;

I'd imagine that for the most part, we don't really have any large images where this is going to make an ultra large difference, but, it exists and we might as well use it?

Strokkur424:
Ah, thank you for the explanation! In that case I don't see a reason to not do that, yeah

zlataovce (approved):

PR #620, opened 2025-07-05
chore: update pdc docs

Timongcraft:
- Refactor examples to use ItemStack instead of ItemMeta
- Small Cleanup


src/content/docs/paper/dev/api/pdc.md
```diff
@@ -49,8 +61,9 @@ item.editMeta(meta -> {
 It is considered good practice to reuse `NamespacedKey` objects. They can be constructed with either:
 - A [`Plugin`](jd:paper:org.bukkit.plugin.Plugin) instance and a [`String`](jd:java:java.lang.String) identifier
 - A [`String`](jd:java:java.lang.String) namespace and a [`String`](jd:java:java.lang.String) identifier
+- Some classes such as ItemStack and OfflinePlayer provide a read-only container "view" and other than ItemStack, OfflinePlayer has no other method to add data.
 
-The first option is often preferred as it will automatically use the plugin's namespace; however, the second option can be used if you
+The first option is often preferred as it will automatically use the plugin's name as namespace; however, the second option can be used if you
```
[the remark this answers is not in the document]
Timongcraft: no...it will use the name of the plugin as namespace. plugins don't have a key or namespace.

src/content/docs/paper/dev/api/pdc.md
```diff
@@ -37,10 +37,22 @@ which is the object you want to store the data on. The third is the data itself.
...
+
+PersistentDataContainer pdc = world.getPersistentDataContainer();
+
+pdc.set(key, PersistentDataType.STRING, "I love Tacos!");
+```
+
+[`ItemStack`](jd:paper:org.bukkit.inventory.ItemStack) however doesn't have this method and instead requires you to use it's builder-style consumer:
+
+```java
+NamespacedKey key = ...; // Retrieve the key from before
```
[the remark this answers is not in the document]
zlataovce: from the previous code block

src/content/docs/paper/dev/api/pdc.md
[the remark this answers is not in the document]
Strokkur424: It is still referred to as the plugin's namespace. If it used the plugin's name, Adventure would throw an exception because of it not being lowercased (as most plugins' names contain uppercase characters)

src/content/docs/paper/dev/api/pdc.md
```diff
@@ -178,6 +198,19 @@ E.g. Placing an ItemStack as a Block (with a TileState) ***does not*** copy over
 Objects that can have a PDC implement the [`PersistentDataHolder`](jd:paper:org.bukkit.persistence.PersistentDataHolder) interface
 and their PDC can be fetched with [`PersistentDataHolder#getPersistentDataContainer()`](jd:paper:org.bukkit.persistence.PersistentDataHolder#getPersistentDataContainer()).
 
+- ##### [`ItemStack`](jd:paper:org.bukkit.inventory.ItemStack)
+    - The persistent data container of an `ItemStack` has historically been accessed by
+      the `ItemStack`'s `ItemMeta`. This, however, includes the overhead of constructing the entire `ItemMeta`, which acts as a snapshot of the `ItemStack`'s data at the point of creation.
+
+      To avoid this overhead, ItemStack exposes a read-only view of its persistent data container at `ItemStack#getPersistentDataContainer()`.
```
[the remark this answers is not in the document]
Timongcraft: I did not change this but moved it. Why rename it?

src/content/docs/paper/dev/api/pdc.md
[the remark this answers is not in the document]
Timongcraft: this just copies the existing layout

src/content/docs/paper/dev/api/pdc.md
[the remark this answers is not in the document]
Strokkur424: Because I think it would be better phrased this way.

src/content/docs/paper/dev/api/pdc.md
```diff
@@ -124,7 +137,14 @@ The `PersistentDataType`'s job is to "deconstruct" a complex data type into some
...
 
 ```java title="UUIDDataType.java"
+@NullMarked
 public class UUIDDataType implements PersistentDataType<byte[], UUID> {
+
+    public static final UUIDDataType INSTANCE = new UUIDDataType();
+
+    // We just need a singleton, so there's no need to allow instantiation
+    private UUIDDataType() {}
+
```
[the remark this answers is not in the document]
Strokkur424: Well aware, just not a particularly big fan of the ps INSTANCE

src/content/docs/paper/dev/api/pdc.md
```diff
@@ -49,8 +61,9 @@ item.editMeta(meta -> {
 It is considered good practice to reuse `NamespacedKey` objects. They can be constructed with either:
 - A [`Plugin`](jd:paper:org.bukkit.plugin.Plugin) instance and a [`String`](jd:java:java.lang.String) identifier
 - A [`String`](jd:java:java.lang.String) namespace and a [`String`](jd:java:java.lang.String) identifier
+- Some classes such as ItemStack and OfflinePlayer provide a read-only container "view" and other than ItemStack, OfflinePlayer has no other method to add data.
```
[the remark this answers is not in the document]
Timongcraft: you can't edit an OfflinePlayer, that's just it.
do you maybe want to propose an improved version?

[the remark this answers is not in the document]
Strokkur424: Maybe this instead?

Certain classes, like `ItemStack` or `OfflinePlayer`, provide a read-only view of their PDC.
In contrast to `ItemStack`, `OfflinePlayer` does **not** provide ways to modify the underlying PDC.

Feel free to write out PDC, I am too lazy for that xd

Timongcraft:
> In general I think you could provide more JavaDoc links here and there

As I see it the docs only link to stuff one timer per document and I think that is fine

src/content/docs/paper/dev/api/pdc.md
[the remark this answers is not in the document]
Strokkur424: Yeah sure

Timongcraft:
ahhh, you can't use double underline?

src/content/docs/paper/dev/api/pdc.md
```diff
@@ -2,21 +2,22 @@
 title: Persistent data container (PDC)
 description: A guide to the PDC API for storing data.
 slug: paper/dev/pdc
+version: "Up to 1.21.5"
```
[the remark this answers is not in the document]
Timongcraft: yeah, I saw that and didn't know how to improve it. good idea!

src/content/docs/paper/dev/api/pdc.md
```diff
@@ -37,20 +38,37 @@ which is the object you want to store the data on. The third is the data itself.
...
+Note: The [`ItemStack#editPersistentDataContainer()`](jd:org.bukkit.inventory.ItemStack#editPersistentDataContainer()) method on `ItemStack` is only available in 1.21.4+. For older versions, you need to access and modify the `ItemMeta` instead.
+For 1.16.5+, there's the [`ItemStack#editMeta()`](jd:org.bukkit.inventory.ItemStack#editMeta(Consumer)) method though.
 
 :::note
 
 It is considered good practice to reuse `NamespacedKey` objects. They can be constructed with either:
 - A [`Plugin`](jd:paper:org.bukkit.plugin.Plugin) instance and a [`String`](jd:java:java.lang.String) identifier
 - A [`String`](jd:java:java.lang.String) namespace and a [`String`](jd:java:java.lang.String) identifier
+- Certain classes, like `ItemStack` or [`OfflinePlayer`](jd:paper:org.bukkit.OfflinePlayer), provide a read-only view of their PDC.
+In contrast to `ItemStack`, `OfflinePlayer` does <u>not</u> provide any way to modify the underlying container.
```
Emilxyz, line 69: This seems out of place in a listing on how to create keys. Especially since the next sentence goes back to creating keys again. So it feels kind of awkward to read jumping from creating keys -> ItemStack & OfflinePlayer -> creating keys.
Maybe it would be better as a paragraph after all the key creation stuff, or in a separate note entirely? Or just mentioning it in the related sections in `Storing on different objects`?

Timongcraft: maybe like this?
https://timongcraft-chore-update-pdc.papermc-docs.pages.dev/paper/dev/pdc/#read-only-containers

Emilxyz: Yeah, this is better.

src/content/docs/paper/dev/api/pdc.md
```diff
@@ -37,20 +38,35 @@ which is the object you want to store the data on. The third is the data itself.
...
+
+// For 1.20.4 and below, use 'new ItemStack(Material.DIAMOND)' instead
+ItemStack item = ItemType.DIAMOND.createItemStack();
+// ItemStack provides a util method, so we can directly edit its PDC
+item.editPersistentDataContainer(pdc -> {
+    pdc.set(key, PersistentDataType.STRING, "I love Tacos!");
 });
 ```
+Note: The [`ItemStack#editPersistentDataContainer()`](jd:org.bukkit.inventory.ItemStack#editPersistentDataContainer()) method on `ItemStack` is only available in 1.21.4+. For older versions, you need to access and modify the `ItemMeta` instead.
+For 1.16.5+, there's the [`ItemStack#editMeta()`](jd:org.bukkit.inventory.ItemStack#editMeta(Consumer)) method though.
```
zlataovce, line 61: use an aside/admonition https://starlight.astro.build/guides/authoring-content/#note-aside

the Javadoc links are invalid, because they do not specify a project
```suggestion
:::note

The [`ItemStack#editPersistentDataContainer()`](jd:paper:org.bukkit.inventory.ItemStack#editPersistentDataContainer()) method on `ItemStack` is only available in 1.21.4+. For older versions, you need to access and modify the `ItemMeta` instead.
For 1.16.5+, there's the [`ItemStack#editMeta()`](jd:paper:org.bukkit.inventory.ItemStack#editMeta(Consumer)) method though.

:::
```

src/content/docs/paper/dev/api/pdc.md
```diff
@@ -63,14 +79,15 @@ To get data from the PDC, you need to know the `NamespacedKey` and the `Persiste
...
-    String value = container.get(key, PersistentDataType.STRING);
+// Get the data from the PDC. Do note that ItemStack provides a "view", which is read-only
+PersistentDataContainerView containerView = item.getPersistentDataContainer();
+if (containerView.has(key, PersistentDataType.STRING)) {
+    String value = containerView.get(key, PersistentDataType.STRING);
     // Do something with the value
     player.sendMessage(Component.text(value));
 }
 ```
+Note: The [`ItemStack#getPersistentDataContainer()`](jd:org.bukkit.inventory.ItemStack#getPersistentDataContainer()) method on `ItemStack` is only available in 1.21.1+. For older versions, you need to access and modify the `ItemMeta` instead.
```
zlataovce, line 90: same here

src/content/docs/paper/dev/api/pdc.md
```diff
@@ -160,11 +184,20 @@ In order to use your own `PersistentDataType`, you must pass an instance of it t
...
 [`has`](jd:paper:io.papermc.paper.persistence.PersistentDataContainerView#has(org.bukkit.NamespacedKey,org.bukkit.persistence.PersistentDataType)) methods.
 ```java
-container.set(key, new UUIDDataType(), uuid);
+container.set(key, UUIDDataType.INSTANCE, uuid);
 ```
 
 :::
 
+
+## Read-only containers
```
zlataovce, line 193: this should be just an aside
```suggestion
```

zlataovce (changes requested):

src/content/docs/paper/dev/api/pdc.md
zlataovce: that is ok

src/content/docs/paper/dev/api/pdc.md
Timongcraft: No, because then it wouldn't be in the sidebar

zlataovce: if it's supposed to be a full-fledged section, it should elaborate on the purpose and benefit of accessing a view instead of a mutable version in general, not just a single note that duplicates information from the "Storing on different objects" section

Timongcraft: > it should elaborate on the purpose and benefit of accessing a view instead of a mutable version in general
What should I elaborate on (other than not accidentally modifying the data)?
Also there's no "benefit", it is just that you can't get a normal container from e.g. an OfflinePlayer.

zlataovce: well currently the purpose of the view is to side-step technical limitations, like the ItemMeta snapshot costliness or the file I/O intensity of having a mutable offline player PDC, which is pretty noteworthy

src/content/docs/paper/dev/api/pdc.md
```diff
@@ -178,6 +224,19 @@ E.g. Placing an ItemStack as a Block (with a TileState) ***does not*** copy over
 Objects that can have a PDC implement the [`PersistentDataHolder`](jd:paper:org.bukkit.persistence.PersistentDataHolder) interface
 and their PDC can be fetched with [`PersistentDataHolder#getPersistentDataContainer()`](jd:paper:org.bukkit.persistence.PersistentDataHolder#getPersistentDataContainer()).
 
+- ##### [`ItemStack`](jd:paper:org.bukkit.inventory.ItemStack)
```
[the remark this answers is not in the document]
Lulu13022002: I don't mind keeping the compat info, but it feels weird to show the view example before explaining it later in the read only section. That part doesn't need to be a note too.

Timongcraft:
I also changed the message sending in the examples, and I think we don't need a notice about Bungee components in older versions.

src/content/docs/paper/dev/api/pdc.md
```diff
@@ -34,41 +35,58 @@ which is used to identify the data. The second is a [`PersistentDataContainer`](
...
 
-ItemStack item = ...; // Retrieve the item from before
-// Get the data from the PDC
-PersistentDataContainer container = item.getItemMeta().getPersistentDataContainer();
-if (container.has(key, PersistentDataType.STRING)) {
-    String value = container.get(key, PersistentDataType.STRING);
+PersistentDataContainer pdc = world.getPersistentDataContainer();
+// Utilize the data from the PDC
+if (pdc.has(key, PersistentDataType.STRING)) {
+    String value = pdc.getOrDefault(key, PersistentDataType.STRING, "<null>");
```
Strokkur424: The getOrDefault is unnecessary here because you are already checking whether the key exists

src/content/docs/paper/dev/api/pdc.md
```diff
@@ -2,21 +2,22 @@
 title: Persistent data container (PDC)
 description: A guide to the PDC API for storing data.
 slug: paper/dev/pdc
+version: "1.21.4 and below"
```
Strokkur424, line 5: I dislike this. This reads like the page is only accurate for the version 1.21.4 and below, but not for 1.21.5+.

Strokkur424 (changes requested):

src/content/docs/paper/dev/api/pdc.md
Timongcraft: Yea, I did that because 1.21.5 was never de-experimentalized and the 1.21.6/7/8 was unstable, could probably bump now

Timongcraft: they where experimental, and thus could still change the api

PR #626, opened 2025-07-20
update packet limiter override

Machine-Maker:


Strokkur424 (changes requested):
image

I don't think the "" being shown makes this intuitive. Maybe also copy over the three setable values with some randomly chosen defaults in order to provide a better preset for copying?

Strokkur424 (approved):

PR #631, opened 2025-08-05
feat: Add datapack discovery lifecycle API docs

### LINKED ISSUE #430: Document datapack registration lifecycle event
https://github.com/PaperMC/Paper/pull/11804

Strokkur424:
Closes #430


src/content/docs/paper/dev/api/lifecycle-api/datapack-registration.mdx
```diff
@@ -0,0 +1,177 @@
...
+```
+
+:::tip
+
+You should always combine `Class#getResource` calls with null checks. Here, we do a simple `Objects.requireNonNull`,
+but you can extend this to for more sophisticated logging and error handling.
+
+:::
+
+### Verifying that the datapack correctly loaded
```
Doc94, line 150: i think you can already check that in the discoverPack step too no?

Strokkur424: You cannot, no. The object you get is only a [`DiscoveredDatapack`](https://jd.papermc.io/paper/1.21.8/io/papermc/paper/datapack/DiscoveredDatapack.html), which doesn't hold any kind of information about its loaded status. I mean, why would it? This is not only about packs which *will* attempt to load (`isRequired()`), but also about if the datapack has a syntax error along the way, which is not checked yet, as they are not loaded yet.

src/content/docs/paper/dev/api/lifecycle-api/datapack-registration.mdx
```diff
@@ -0,0 +1,177 @@
...
+
+Alternatively, you can check for the loaded status of your datapack during normal execution
+of your plugin. For example, a simple check inside your plugin's `onLoad` method might look like this:
+
+```java title="CustomJavaPlugin.java"
+public final class CustomJavaPlugin extends JavaPlugin {
+
+    @Override
+    public void onLoad() {
+        Datapack pack = Bukkit.getServer().getDatapackManager().getPack(getPluginMeta().getName() + "/provided");
```
Lulu13022002, line 161: Get the server from your plugin (this#getServer)

src/content/docs/paper/dev/api/lifecycle-api/datapack-registration.mdx
```diff
@@ -0,0 +1,177 @@
...
+        if (pack != null) {
+            if (pack.isEnabled()) {
+                getComponentLogger().info("The datapack loaded successfully!");
```
Lulu13022002, line 164: Any reason to use the component logger for plain message?

src/content/docs/paper/dev/api/lifecycle-api/datapack-registration.mdx
```diff
@@ -0,0 +1,177 @@
...
+
+The lifecycle API grants developers much more direct access to modifying some of the core parts of the server.
+One such core aspect are **datapacks**. No more asking users to download your datapack alongside the plugin - you
+can now include the datapack in your plugin JAR and load it yourself!
+
+:::note
+
+This feature requires you to have a basic understanding of the [Lifecycle API](/paper/dev/lifecycle),
+the datapack format (which you can look up in the [Minecraft wiki](https://minecraft.wiki/w/Data_pack)), and
+requires you to use a [`paper-plugin.yml` plugin](/paper/dev/getting-started/paper-plugins).
```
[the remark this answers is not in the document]
Strokkur424: What's the alternative though? Calling it "Paper plugin" is confusing, and the migration to "lifecycle plugins" is not at all official. Whilst it does look a bit weird, this is the most unconfusing phrasing I could come up with.

src/content/docs/paper/dev/api/lifecycle-api/datapack-registration.mdx
Strokkur424: I've never really thought about it much, but does it actually make a difference in any way? (I will change it, but I am just curious).

src/content/docs/paper/dev/api/lifecycle-api/datapack-registration.mdx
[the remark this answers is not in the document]
Lulu13022002: Right, ignore this i think there was a whole discussion about renaming "paper plugins" and all, i personally disagree with the lifecycle plugins term too.

src/content/docs/paper/dev/api/lifecycle-api/datapack-registration.mdx
```diff
@@ -0,0 +1,183 @@
...
+          - pack.mcmeta
+          - data/
+</FileTree>
+
+Build your plugin and verify that there is a `custom_datapack` folder in the root of your plugin's JAR file.
+
+### Discovering the datapack
+In order to discover the datapack, you must call the [`DatapackRegistrar#discoverPack(URI uri, String id)`](jd:paper:io.papermc.paper.datapack.DatapackRegistrar#discoverPack(java.net.URI,java.lang.String))
+method. The uri should point to **your datapack's folder in your JAR**. This can be achieved simply by calling
+`getClass().getResource("/custom_datapack").toURI()`. **The trailing slash is very important**. The id can be set to
```
Machine-Maker, line 132: Not a trailing slash. Its a preceding slash.

Machine-Maker (changes requested):

src/content/docs/paper/dev/api/lifecycle-api/datapack-registration.mdx
```diff
@@ -0,0 +1,183 @@
...
+    // that your plugin includes this data pack (as the name is prefixes with the plugin name).
+    event.registrar().discoverPack(uri, "provided");
+} catch (URISyntaxException | IOException e) {
+    throw new RuntimeException(e);
+}
+```
+
+:::tip
+
+You should always combine `Class#getResource` calls with null checks. Here, we do a simple `Objects.requireNonNull`,
```
zlataovce, line 151: there's a reasonable expectation that a static classpath resource, such as a datapack or plugin config file, will exist, worst case it fails with a NPE at toURI, so why is this specially pointed out

Strokkur424: Generally just because it is good practice to try and catch NPEs as soon as possible so that you reduce the time it takes for you to debug something *if* something should go wrong anyways.
Just imagine you do the call at the start of a method, then you do some unrelated work, and only then you try to `toURI` it. That's really all there is to it.

Do you think it is unnecessary and should I remove it? I was just wanted to promote good practice.

src/content/docs/paper/dev/api/lifecycle-api/datapack-registration.mdx
```diff
@@ -0,0 +1,178 @@
...
+                this.getLogger().info("The datapack loaded successfully!");
+            } else {
+                this.getLogger().warn("The datapack failed to load.");
+            }
+        }
+    }
+}
+```
+
+If everything has gone correctly, the console should contain input similar to this:
```
Privatech38, line 174: I would reword this to something in the lines of "should output something similar to:"

PR #634, opened 2025-08-08
feat: Improve basic commands page

Strokkur424:
- Improves the basic commands page by suggesting the simpler way of registering commands.
- Adds a tip aside in the introductory page to slightly advertise their use for people not wanting to jump into the hell that is Brigadier.


stivendarsi:
Add in big scary text that says Brigadier is better for a more sorted command tree

Issue #645, opened 2025-08-20
Velocity docs on events show outdated values


stivendarsi:
<img width="821" height="306" alt="Image" src="https://github.com/user-attachments/assets/e200fec3-f514-4292-9be6-8bf1a3be2786" />

<img width="779" height="182" alt="Image" src="https://github.com/user-attachments/assets/67572158-a05f-4b8e-a7a5-0b9ebcd8cf31" />

PR #646, opened 2025-08-20
feat: update velocity docs for events priority

Doc94:
This PR close https://github.com/PaperMC/docs/issues/645 where the order/priority docs for events are outdated with the current behaviour

- PostOrder.CUSTOM not necesary since https://github.com/PaperMC/Velocity/commit/7392cd6574fcc5b1496fe4c3ac2b6fa7c8e993d7
- order field for Subscribe deprecated since https://github.com/PaperMC/Velocity/commit/4f227badc20dc30b0f6d84b5349c8809481dcbb1

### DIFF
--- a/src/content/docs/velocity/dev/api/event.md
+++ b/src/content/docs/velocity/dev/api/event.md
@@ -40,12 +40,12 @@ State the desired order in the `@Subscribe` annotation:
 ```java
 @Subscribe(priority = 10)
 public void onPlayerChatFirst(PlayerChatEvent event) {
-	// do first stuff
+    // this method fires first
 }
 
 @Subscribe(priority = 5)
 public void onPlayerChat(PlayerChatEvent event) {
-  // do other stuff with the change from the first listener
+    // this method fires after the one above
 }
 ```
 


src/content/docs/velocity/dev/api/event.md
```diff
@@ -38,18 +38,13 @@ The higher the priority, the earlier the event handler will be called.
...
 
 ```java
-@Subscribe(priority = 0, order = PostOrder.CUSTOM)
+@Subscribe(priority = 0)
 public void onPlayerChat(PlayerChatEvent event) {
 	// do stuff
 }
 ```
 
 `-32768` is the default value if you do not specify an order.
```
Strokkur424, line 47: Two things:

1. Looking at the source code, it seems that `0` is the new default `priority`. Could you maybe change the value here to reflect that?

2. More generally, I think we should rephrase this part around here a bit, considering in this example we don't use any order at all, rather we should refer to it as priority. (Also true for L38).

src/content/docs/velocity/dev/api/event.md
```diff
@@ -92,7 +87,7 @@ public class VelocityTest {
 
 public class MyListener {
 
-  @Subscribe(order = PostOrder.EARLY)
+  @Subscribe(priority = 0)
```
Strokkur424, line 90: Seeing as the priority is `0` to default, could we not just remove the explicit mapping here entirely?

Strokkur424 (changes requested):

PR #652, opened 2025-09-12
feat: document the /paper command

### LINKED ISSUE #650: Add Paper built-in commands
It would be nice to add a section for the Paper built-in commands with a description and/or an explanation of the output, for example /paper chunkinfo - explain the usage and the output (what are inactive/total/full etc)

Strokkur424:
Closes #650.


zlataovce:
I think the page should be generalized to commands, because there are also other noteworthy commands inherited from Spigot

PR #658, opened 2025-09-25
fix: small issue in recipe docs

m4sc0:
Updated diamond sword crafting recipe details.

### DIFF
--- a/src/content/docs/paper/dev/api/recipes.md
+++ b/src/content/docs/paper/dev/api/recipes.md
@@ -21,7 +21,7 @@ public class TestPlugin extends JavaPlugin {
         NamespacedKey key = new NamespacedKey(this, "television");
 
         ItemStack item = ItemStack.of(Material.BLACK_WOOL);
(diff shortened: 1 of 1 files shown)


m4sc0:
omg i did it wrong...

Strokkur424 (changes requested):
For once there is an issue with the original, which this PR is trying to fix, where this almost seems like an accident in the sword recipe. But then again, as it was pointed out, remaking a Vanilla recipe is maybe a bit weird. Thereby I suggest that we maybe find a more custom recipe. Could be something simple like a "Television" with like 8 white concrete surrounding a piece of black stained glass pane and out comes a piece of wool called "Television". I am not the most creative person, but you get the idea. I think that would clear up any misconceptions here. *(Please, feel free to have your own idea here.)*

Furthermore, as lulu already pointed out, this segment here is worded weirdly:
> The diamond sword would be in the middle column of
the bottom row.

I am not entirely certain what this was meant to describe, but seeing as, if you remove that sentence entirely, it would still make sense, I think we can just omit it.

PR #671, opened 2025-10-06
Fix community library links

Siebrenvde:
These links lead to a 404

Was also mentioned here but seems to have been forgotten: https://github.com/PaperMC/docs/pull/573#discussion_r2097747453


Strokkur424:
Thank you! That was indeed an oversight.

PR #678, opened 2025-10-15
Adventure 5.0

kezz:
**To-do list**
- [x] Add a note about abstract component renderer changes
- [x] Add a note about text format being sealed
- [x] Add a change entry about the GSON library version bump (we will try not to use new features, also the quirk about `null` now returning `null` and not throwing an exception)
- [x] Document the removal of deprecated JSON serializer methods that are now done via options instead
- [x] Add a note about the switch to JSpecify annotations
- [x] Add a note about custom click payload nbt field becoming nullable
- [x] Add a note about the new-nullability of ChatType#key and the fact it doesn't implement Keyed anymore
- [x] Document the removal of UTF8ResourceBundleControl
- [x] Nullability isn't at module level, it's class/package level
- [x] Remove the SBA synthetic stuff
- [x] Give the rest of the docs a quick look over to ensure we're not yapping/exampling anything that's now removed


src/content/docs/adventure/migration/adventure-4.x.md
```diff
@@ -0,0 +1,104 @@
...
+
+With the release of Adventure 5.0, some breaking changes have been introduced from the Adventure 4.x series.
+This page documents the changes made and how you as a developer can migrate your code.
+
+## A modern codebase
+One of the main goals for Adventure 5.0 was to migrate to a more modern codebase.
+The minimum version of Java required to use Adventure is now Java 21.
+
+By updating to Java 21, Adventure has taken advantage of sealed classes and interfaces.
+Almost every interface/class that was annotated with `@ApiStatus.Internal` has now been made sealed.
```
kezz:
```suggestion
Almost every interface/class that was annotated with `@ApiStatus.NonExtendable` has now been made sealed.
```

Strokkur424 (approved):

zlataovce (approved):

Privatech38:
This PR should also adress https://github.com/PaperMC/docs/issues/749 since v5 throws an error if SNBT is used instead of a list of data components

kezz:
> This PR should also adress #749 since v5 throws an error if SNBT is used instead of a list of data components

If you are aware of issues or bugs in Adventure, please report them there.

PR #680, opened 2025-10-18
Add source links for dynamic replacement examples

JavierFlores09:


Doc94:
Not sure about use static reference to github

JavierFlores09:
> Not sure about use static reference to github

I was considering permalinks but it would get outdated if they were to change the classes, if you're fine with that then I'll just change those into permalinks. Should be fine unless they were to move the class around or change the branch name honestly

zlataovce (approved):

PR #686, opened 2025-11-07
feat(adventure): add MiniMessage language specification

Strokkur424:


src/content/docs/adventure/minimessage/specification.md
```diff
@@ -0,0 +1,314 @@
...
+```
+
+Tags can be closed by repeating the tag, with a slash in front of the name. Tags are closed implicitly
+when the end of the string is reached. Furthermore, tags can be nested:
+
+```mm
+<first_tag>Some text <second_tag>even more text</second_tag>, and that's really it!
+```
+
+Nested tags are closed implicitly when the outer tag is closed.
```
Andre601, line 75: This sentence seems inaccurate from my own experience.

When f.e. using a color tag inside a formatting tag, closing said formatting tag doesn't automatically close the color tag, as this line here would imply.
What is true, is that any open tags will be closed when the end of a String is reached and tags remain open.

PR #688, opened 2025-11-13
feat: generate Javadoc link text automatically

zlataovce:
Got a little frustrated over the inconsistency of these, so added a way to have it be generated from the reference.

It is demoed on the "Chat events" page.

### DIFF
--- a/src/content/docs/paper/dev/api/event-api/chat-event.md
+++ b/src/content/docs/paper/dev/api/event-api/chat-event.md
@@ -39,7 +39,7 @@ public Component render(Player source, Component sourceDisplayName, Component me
 }
 ```
 
-- The [](jd:paper:io.papermc.paper.chat.ChatRenderer#render(org.bukkit.entity.Player,net.kyori.adventure.text.Component,net.kyori.adventure.text.Component,net.kyori.adventure.audience.Audience)) method is called when a chat message is sent to the player.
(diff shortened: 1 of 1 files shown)


src/content/docs/paper/dev/api/event-api/chat-event.md
```diff
@@ -28,18 +28,18 @@ However, we recommend using [`BukkitScheduler`](/paper/dev/scheduler).
...
+with its [](jd:paper:org.bukkit.event.player.AsyncPlayerChatEvent#setFormat(java.lang.String)) method.
 
 ```java title="ChatRenderer#render"
 public Component render(Player source, Component sourceDisplayName, Component message, Audience viewer) {
     // ...
 }
 ```
 
-- The [`render`](jd:paper:io.papermc.paper.chat.ChatRenderer#render(org.bukkit.entity.Player,net.kyori.adventure.text.Component,net.kyori.adventure.text.Component,net.kyori.adventure.audience.Audience)) method is called when a chat message is sent to the player.
+- The [](jd:paper:io.papermc.paper.chat.ChatRenderer#render(org.bukkit.entity.Player,net.kyori.adventure.text.Component,net.kyori.adventure.text.Component,net.kyori.adventure.audience.Audience)) method is called when a chat message is sent to the player.
```
Strokkur424: I think here just showing `render` without the parameters is cleaner, as the method is shown above and the params are explained below

PR #691, opened 2025-11-19
Expand PreparedStatement section and add an entry about SQL tools/libraries

Biquaternions:
Added a section that would've been incredibly useful when I started and will for sure be as well for new developers.
This PR adds a small section talking about lightweight tools like JDBI and JOOQ to handle SQL statements in plugins.

For comparison, I also added an example on the `PreparedStatement` section showing how the example code provided in the `Security` would look like using it.

To prevent bloating the section, I avoided any unnecessary details (at least to my criteria), added links to the documentation of the two mentioned projects and minimized the code examples (like not using try/catch statements)

### DIFF
--- a/src/content/docs/paper/dev/misc/databases.md
+++ b/src/content/docs/paper/dev/misc/databases.md
@@ -211,26 +211,9 @@ public void login(DataSource dataSource, String username, String password) {
 
 ## Database tools
 
-A common approach to handle SQL security and other concerns like type-safety, is to use a library that can take care of the heavy lifting,
-leaving the developer with clean, secure and readable code.
+Given the complexity of working with databases (managing connections, building and securing queries, or just parsing the data) several tools
+exist in the world of Java to leverage this work.
 
-Libraries can take multiple approaches to handle data, like SQL builders or [ORMs](https://en.wikipedia.org/wiki/Object%E2%80%93relational_mapping). \
-Some of the most common libraries include [JDBI](https://jdbi.org/) and [JOOQ](https://www.jooq.org/doc/latest/manual/).
-
-For example, the following SQL statement:
-
-```sql
-SELECT * FROM users WHERE username = '' AND password = '';
-```
-
-Will become the following code in Java using JOOQ:
-
-```java
-public void login(DSLContext context, String username, String password) {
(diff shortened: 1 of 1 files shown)


src/content/docs/paper/dev/misc/databases.md
```diff
@@ -192,3 +192,41 @@ Using prepared statements in Java with [`PreparedStatement`](jd:java:java.sql:ja
 helps prevent SQL injection. They separate SQL code from user input by using placeholders, reducing the risk of executing unintended SQL commands.
 **Always** use prepared statements to ensure the security and integrity of your data. Read more about SQL injection
 [here](https://www.baeldung.com/sql-injection).
+
+When using `PreparedStatement` the `login` method will become:
+
+```java
+public void login(String username, String password) throws SQLException {
```
electronicboy: if we're handing out code examples of stuff this really should follow the safe standard practices, such as using TWR to ensure that we clean up

src/content/docs/paper/dev/misc/databases.md
```diff
@@ -192,3 +192,41 @@ Using prepared statements in Java with [`PreparedStatement`](jd:java:java.sql:ja
...
+```sql
+SELECT * FROM users WHERE username = '' AND password = '';
+```
+
+Will become the following code in Java using JOOQ:
+
+```java
+public void login(String username, String password) {
+    UsersRecord user = this.jooq.selectFrom(USERS)
+            .set(USERS.USERNAME, username)
```
electronicboy: Not ultra familiar with jooq but this doesn't appear to be valid?

Biquaternions: It is valid, but since I tried to keep the code as minimal as possible, I added a `this.jooq` variable and called it a day.
Given your other comment about using real appropriate examples, I will expand this one as well.

For reference I'm using this project I recently updated from plain jdbc to jooq:
https://github.com/SerlithNetwork/Nascraft/blob/master/src/main/java/me/bounser/nascraft/database/commands/PlayerStats.java#L48

Biquaternions: I just saw what you meant, I think I copied from insertInto while building the example

PR #717, opened 2026-01-19
chore: better show Velocity's async handling

Timongcraft:
some users (2) came confused in the velocity channels because they didn't understand this example so this should be more self-explanatory


Timongcraft:
The thing is that the previous example used code used an external lib not well known or was supposed to represent a concept that some users didn't understand this makes it clear and also shows how the resume methods are used.

Also the _ will be highlighted by Intellij if it isn't possible so that shouldn't be a problem. (This is Java 22+, Velocity is Java 17)
Can change this though if desired

PR #724, opened 2026-02-02
Add Imperat library to community libraries list

iiAhmedYT:


iiAhmedYT:
> The Adventure community libraries section is supposed to be used for libraries made by the community **which directly utilize Adventure**. It is not a place to put your command framework.

it does indeed utilize adventure and has direct support for it on Velocity, Bukkit and Bungeecord

iiAhmedYT:
@kashike Would you be so kind to provide a reason to why this was closed?

PR #735, opened 2026-02-21
feat(velocity): document velocity-plugin.json file

### LINKED ISSUE #731: Document the Velocity plugin JSON file format
The annotation processor, which is currently recommended to be used, is not exactly the nicest to use for anything slightly more dynamic (since variable substitution is hard). For this reason, we should document the Velocity plugin JSON file in order to make the manual inclusion possible.

I also think we should remove (or at least inline) any mentions of the AP for generating that file and instead make manually writing the file the preferred way.

Strokkur424:
closes #731

I actually decided against outright removing mentions of the AP, as it *is* useful for beginners and honestly doesn't really hurt most people. But having the actual file documented seems like a good thing.


src/content/docs/velocity/dev/getting-started/velocity-plugin-json.mdx
```diff
@@ -0,0 +1,90 @@
...
+### authors
+A list of your plugin's authors.
+- `"authors": ["Strokkur24", "electronicboy"]`
+
+### dependencies
+Your plugin's dependencies. The value of this field is an array of dependency objects.
+
+The dependency object itself has three fields:
+- `id` the plugin ID of the plugin you depend upon
+- `version` the version of your dependency. This field is not required.
```
Emilxyz: I don't think this exists in the current version of velocity:
https://github.com/PaperMC/Velocity/blob/7d0c002f89d1cc916fd7525bcb4f2a0f4bc31649/api/src/ap/java/com/velocitypowered/api/plugin/ap/SerializedPluginDescription.java#L207
```suggestion
```

src/content/docs/velocity/dev/getting-started/creating-your-first-plugin.mdx
```diff
@@ -57,6 +57,15 @@ How to set up a build system is outside the scope of this page, but you can look
 system's documentation ([Gradle](https://docs.gradle.org/current/userguide/userguide.html) or
 [Maven](https://maven.apache.org/guides/getting-started/index.html)) for assistance.
 
+:::info[velocity-plugin.json]
```
Timongcraft: Looks like this doesn't work?

Image

qheilmann:
> closes #731
> 
> I actually decided against outright removing mentions of the AP, as it _is_ useful for beginners and honestly doesn't really hurt most people. But having the actual file documented seems like a good thing.

I'm grateful I began exploring Velocity yesterday and was pondering, "What exactly is the velocity-plugin.json? Is it similar to the paper-plugin.yml, so what is the purpose of the @Plugin annotation?"
This PR provides me all the information needed, thanks !

Zoriot:
It seems like either "Id" or "Name" has to be present for it to be valid. It doesn't have to be "Id"

electronicboy:
Generally looks fine to me, only real thing is the addition of "provides" which takes a list of IDs

PR #767, opened 2026-05-18
feature(adventure): add notes about archival of -platform

kezz:


zlataovce (approved):
could also use a destructive banner like on the Waterfall pages if the asides don't grab attention enough

PR #768, opened 2026-05-30
Typo: "Support's" should be "Supports"

Cubicake:
"Support's" is not grammatically correct, "supports" should be used.

Just a single character fix I noticed when reading the page.


zlataovce:
please fix it also for the other occurrences, otherwise LGTM

Cubicake:
Oh right that message is duplicated for all unmaintained platforms. I think that's all of them.

zlataovce (approved):

PR #775, opened 2026-06-25
chore(adventure): replace JD URL links with markdown syntax

Doc94:
With the migration of the adventure javadocs to https://jd.papermc.io/ this PR manage to use the jd component for point to that docs and also update the javadoc url in the adventure section (used for the j icon)


astro.config.ts
```diff
@@ -569,6 +569,7 @@ export default defineConfig({
             targets: {
               paper: "https://jd.papermc.io/paper",
               velocity: "https://jd.papermc.io/velocity",
+              adventure: "https://jd.papermc.io/adventure",
```
Doc94: the only "issue" is the whole module reference like `(jd:adventure:net.kyori.adventure.api:net.kyori.adventure.sound.Sound.Emitter#self())` maybe can be good include all the modules like `adventure_api: { url: "https://jd.papermc.io/adventure", module: "net.kyori.adventure.api" },` and then use `(jd:adventure_api:net.kyori.adventure.sound.Sound.Emitter#self())` but its just for have less text in the component....

Strokkur424:
I have no idea why the deployment build fails in CI, as it works perfectly fine locally with `pnpm build`, which is also what the CI does. I wish you the best fixing it, Doc!

electronicboy:
> Error: Received dead status for Javadoc link "jd:adventure:net.kyori.adventure.api:net.kyori.adventure.chat.ChatType.Bound"

I think the CI env runs with some extra checks? either that or the builder is getting refused access to the JDs

Strokkur424:
Oh yeah, appending `CI=true` before the build command does make it run into the same problem. Thanks for the heads-up, cat. Will look if I can figure this out.

Strokkur424 (approved):

PR #782, opened 2026-07-20
feat(misc): Add note to forums for beeing discontinued

Chamogelastos:


Privatech38:
I don't think removing it entirely is the best option here. Instead add a notice that is discontinued. The forums are still up and contain old, but useful threads.

src/content/docs/misc/contact.md
```diff
@@ -10,10 +10,17 @@ The PaperMC project handles most communication via Discord. Use the following in
 
 ## Forums
 
+:::note
+
+After careful consideration and due to limited usage, we’ve made the decision to discontinue the PaperMC forums. Moving forward, we recommend using Hangar for plugin uploads, and for all other community discussions and support, please join us on Discord.
+
+:::
+
```
Privatech38: You don't need whitespace in asides. I also recommend using caution instead.
```suggestion
:::caution
After careful consideration and due to limited usage, we’ve made the decision to discontinue the PaperMC forums. Moving forward, we recommend using Hangar for plugin uploads, and for all other community discussions and support, please join us on Discord.
:::
```

src/content/docs/misc/contact.md
```diff
@@ -10,10 +10,17 @@ The PaperMC project handles most communication via Discord. Use the following in
...
 Reach out for support, or contact us on our forums.
 
 - [Forums](https://forums.papermc.io)
 
+
```
Privatech38: Why the extra whitespace here?
```suggestion
```

Privatech38 (changes requested):

src/content/docs/misc/contact.md
Chamogelastos: No reason at all

src/content/docs/misc/contact.md
Chamogelastos: Should we do it as a "note" or "caution" would be better

Warriorrrr:
> I don't think removing it entirely is the best option here. Instead add a notice that is discontinued. The forums are still up and contain old, but useful threads.

But as a means of getting in contact with anyone, a read-only forum is useless.

src/content/docs/misc/contact.md
Privatech38: I believe this calls for a caution. Note is more useful for stuff like "oh btw this thing does this small thing as well". Caution is more like "hey before you go ahead, this thing does something you should know about".
Danger is "dont do anything before reading this"

Chamogelastos: The only reason I used whitespaces is beause it is used throughout the docs. I also believe it would be more fitting as a note

Doc94:
> > I don't think removing it entirely is the best option here. Instead add a notice that is discontinued. The forums are still up and contain old, but useful threads.
> 
> But as a means of getting in contact with anyone, a read-only forum is useless.

+1 this section is for contact, even not sure if the twitter can be consider a good way to contact....

I remember the website mention a email for contact

Chamogelastos:
I believe the current one is a decent middleground between those two

Doc94:
> I believe the current one is a decent middleground between those two

for me im not sure... Because next to that notice still says you can use for support/contact when its read only and the final of all says the first two are the most check...

Maybe a docs team oppinion can be good

Chamogelastos:
> for me im not sure... Because next to that notice still says you can use for support/contact when its read only and the final of all says the first two are the most check...
> 
> Maybe a docs team oppinion can be good

I mean they always have the last talk

Timongcraft:
Yea IMO the Twitter section should also have a *warning* seeing that the last post is from 2022.
And instead maybe link to the [news](https://papermc.io/news/) page?
But I don't even think that really fits in a 'Contact Us' page

Also there's this idk if it is actively watched though: https://forums.papermc.io/misc/contact

NoahvdAa:
The last twitter post was in May, twitter just hides most posts if you're not signed in.

PR #788, opened 2026-08-05
feat(paper): document scoreboard API

Strokkur424:


src/content/docs/paper/dev/api/scoreboard-api.mdx
```diff
@@ -0,0 +1,359 @@
...
+To set the display slot, you can call [`setDisplaySlot(DisplaySlot)`](jd:paper:org.bukkit.scoreboard.Objective#setDisplaySlot(org.bukkit.scoreboard.DisplaySlot))
+on the [](jd:paper:org.bukkit.scoreboard.Objective) object.
+
+There are three different display slots.
+
+#### Sidebar display slot
+
+The probably most well-known display slot of an objective is the sidebar. That is also what most players generally
+understand as the "scoreboard". It shows up on the right-side of a player's client and consists of a title and
+up to 15 lines of scores. The display information is the same for all players viewing the same scoreboard instance.
```
Leguan16, line 78: This whole sentence sounds weird

Malfrador:
I think a mention that stuff like Fastboard exists would be good. We also mention configurate on the config page for example. 

The API is needlessly complicated and verbose and easily conflicting with other plugins if your only goal is "display some text in the sidebar" because it (rightfully) represents the vanilla scoreboard system with all its features. Which is simply not what you need if you only want a sidebar text display - but thats what most devs are looking for in my experience.

Easily exemplified by the entire 77-line "Sidebars done correctly" example becoming around 5-10 lines of code when using Fastboard instead.

